PR #45, opened 2026-02-01
Fix proxy gzip compression handling for Claude API integration

### LINKED ISSUE #44: Proxy fails to parse Claude API responses due to gzip compression handling
## Problem

The tapes proxy fails to capture Claude Code sessions with the following error:

```
WARN  proxy/proxy.go:191  failed to parse response  
{"error": "invalid character '\\x1f' looking for beginning of value", "provider": "anthropic"}
```

The `\x1f` character is the gzip magic number, indicating the proxy is trying to parse compressed response bytes as JSON.

## Root Cause

In `proxy/proxy.go` lines 170-176, the proxy copies all request headers including `Accept-Encoding`:

```go
// Copy relevant headers
c.Request().Header.VisitAll(func(key, value []byte) {
    k := string(key)
    // Skip hop-by-hop headers
    if k != "Connection" && k != "Host" {
        httpReq.Header.Set(k, string(value))  // ← Copies Accept-Encoding: gzip
    }
})
```

When `Accept-Encoding` is manually set on Go's `http.Client`, it **disables automatic decompression**. The client receives compressed bytes but doesn't decompress them.

## Impact

This breaks Claude Code integration via `ANTHROPIC_BASE_URL`:

```bash
# Fails to capture sessions
export ANTHROPIC_BASE_URL="http://localhost:8080"
tapes serve --upstream "https://api.anthropic.com" --proxy-listen "0.0.0.0:8080"
claude  # Sessions fail to parse
```

## Suggested Fix

Skip the `Accept-Encoding` header and let Go handle compression automatically:

```go
// Copy relevant headers
c.Request().Header.VisitAll(func(key, value []byte) {
    k := string(key)
    // Skip hop-by-hop headers and Accept-Encoding (let Go handle compression)
    if k != "Connection" && k != "Host" && k != "Accept-Encoding" {
        httpReq.Header.Set(k, string(value))
    }
})
```

This allows Go's `http.Client` to:
1. Add `Accept-Encoding: gzip` automatically if the server supports it
2. Auto-decompress responses
3. Return uncompressed bytes to the parsing logic

## Alternative Solutions

1. **Manual decompression**: Check `Content-Encoding` header in response and decompress if needed
2. **Disable compression**: Strip `Accept-Encoding` from client request (forces uncompressed responses)

Option 1 (suggested fix) is cleanest - leverages Go's built-in transparent compression handling.

## Reproduction

1. Start tapes proxy pointing to Anthropic API
2. Configure Claude Code with `ANTHROPIC_BASE_URL=http://localhost:8080`
3. Run `claude` and send a message
4. Observe gzip parsing errors in tapes logs
5. Check `curl http://localhost:8081/dag/stats` - shows 0 nodes captured

## Environment

- tapes version: latest (main branch)
- Go version: 1.25+
- Platform: macOS
- Client: Claude Code CLI

## References

- Go `http.Client` compression docs: https://pkg.go.dev/net/http#Transport
- Related to Claude Code integration via `ANTHROPIC_BASE_URL` pattern

bdougie:
## Summary

Fix proxy gzip compression handling that was breaking Claude API integration by filtering out `Accept-Encoding` headers to let Go handle compression automatically.

Discovered when debugging claude code guide https://github.com/papercomputeco/tapes.dev/pull/25

## Changes

- **proxy/proxy.go:173**: Added `&& k != "Accept-Encoding"` to non-streaming handler header filtering
- **proxy/proxy.go:252**: Added `&& k != "Accept-Encoding"` to streaming handler header filtering  
- **proxy/proxy_test.go**: Added test to verify `Accept-Encoding` header is properly filtered

## Problem

The proxy was copying all client headers including `Accept-Encoding: gzip` to upstream requests. When Go's `http.Client` receives a manually set `Accept-Encoding` header, it disables automatic decompression, causing the proxy to receive compressed gzip bytes and fail JSON parsing with:

```
WARN failed to parse response {"error": "invalid character '\\x1f' looking for beginning of value", "provider": "anthropic"}
```

Blocking the [claude code](https://tapes.dev/guides/claudecode/) and opencode guide. 

## Solution

Filter out the `Accept-Encoding` header and let Go's `http.Client` handle compression transparently:
1. Go automatically adds appropriate `Accept-Encoding` headers
2. Go handles response decompression automatically  
3. Proxy receives uncompressed JSON for parsing

## Impact

- ✅ Fixes Claude Code integration via `ANTHROPIC_BASE_URL`
- ✅ Leverages Go's built-in compression handling
- ✅ Maintains backward compatibility
- ✅ No breaking changes to existing functionality

Fixes #44

### DIFF
--- a/pkg/llm/provider/anthropic/anthropic.go
+++ b/pkg/llm/provider/anthropic/anthropic.go
@@ -3,6 +3,7 @@ package anthropic
 
 import (
 	"encoding/json"
+	"strings"
 	"time"
 
 	"github.com/papercomputeco/tapes/pkg/llm"
@@ -30,6 +31,7 @@ func (p *provider) ParseRequest(payload []byte) (*llm.ChatRequest, error) {
 		return nil, err
 	}
 
+	system := parseAnthropicSystem(req.System)
 	messages := make([]llm.Message, 0, len(req.Messages))
 	for _, msg := range req.Messages {
 		converted := llm.Message{Role: msg.Role}
@@ -78,7 +80,7 @@ func (p *provider) ParseRequest(payload []byte) (*llm.ChatRequest, error) {
 	result := &llm.ChatRequest{
 		Model:       req.Model,
 		Messages:    messages,
-		System:      req.System,
+		System:      system,
 		MaxTokens:   &req.MaxTokens,
 		Temperature: req.Temperature,
 		TopP:        req.TopP,
@@ -89,6 +91,36 @@ func (p *provider) ParseRequest(payload []byte) (*llm.ChatRequest, error) {
 	}
 
 	return result, nil
+}
+
+func parseAnthropicSystem(system any) string {
+	if system == nil {
+		return ""
+	}
+
+	switch value := system.(type) {
+	case string:
+		return value
+	case []any:
+		var builder strings.Builder
+		for _, item := range value {
+			block, ok := item.(map[string]any)
+			if !ok {
+				continue
+			}
+			blockType, _ := block["type"].(string)
+			text, _ := block["text"].(string)
+			if blockType == "text" && text != "" {
+				if builder.Len() > 0 {
+					builder.WriteString("\n")
+				}
+				builder.WriteString(text)
+			}
+		}
+		return builder.String()
+	default:
+		return ""
+	}
 }
 
 func (p *provider) ParseResponse(payload []byte) (*llm.ChatResponse, error) {
--- a/pkg/llm/provider/anthropic/types.go
+++ b/pkg/llm/provider/anthropic/types.go
@@ -4,7 +4,7 @@ package anthropic
 type anthropicRequest struct {
 	Model       string             `json:"model"`
 	Messages    []anthropicMessage `json:"messages"`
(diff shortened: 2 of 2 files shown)


jpmcb (changes requested):
Ah interesting: we shouldn't ignore encodings from clients: some might _only_ accept encodings. I thought this was automatically handled by fiber, but seems we need to add the middleware explicitly:

```go
import (
  "github.com/gofiber/fiber/v2"
  "github.com/gofiber/fiber/v2/middleware/compress"
)

// Init compress middleware
app.Use(compress.New())
```

From their docs: https://docs.gofiber.io/api/middleware/compress

> Compression middleware for [Fiber](https://github.com/gofiber/fiber) that will compress the response using gzip, deflate and brotli compression depending on the [Accept-Encoding](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) header.

bdougie:
Now that the gzip is fixed, I saw the JSON marshaling needed a fix in https://github.com/papercomputeco/tapes/pull/45/changes/b293e114f4480e7387d8d5fa8b7bf6f5c7174dbe

PR #86, opened 2026-02-12
feat: memory Driver interface and "local" implementor

### LINKED ISSUE #81: RFC: tapes memory layer
The current search pipeline is unstructured recall, embed nodes, query by similarity, return branches. It answers "what past conversations are similar to X?" but doesn't synthesize or prioritize.

## What a memory layer could add

A memory config, modeled after the search config pattern, could introduce structured recall:

**Short-term memory**: branch-scoped context
- Recent nodes within the current proxy branch
- Sliding window or token budget (e.g., last N nodes)
- Cheap, fast, ephemeral: lives in the worker pool or in-memory storage driver

**Long-term memory**: persistent, cross-branch knowledge
- [Extracting durable facts/patterns](https://www.cognee.ai/blog/deep-dives/the-art-of-intelligent-retrieval-unlocking-the-power-of-search) from conversations into a knowledge graph
- Goes beyond vector similarity
- Could use a graph store (Cognee-style) or a curated vector collection with metadata

## How it could map to tapes' config pattern

```toml
[memory]
provider = "cognee"       # or "local", "graph"
target = "http://..."

[memory.short_term]
enabled = true
window = 10               # last N nodes in current branch

[memory.long_term]
enabled = true
provider = "cognee"       # knowledge graph extraction
```

This follows the same optional-feature pattern as search:
- **Zero config** — proxy works without it
- **Graceful degradation** — returns "memory not configured" if missing
- **Driver interface** — `memory.Driver` with `Store()`, `Recall()`, `Forget()`
- **Pluggable backends** — local graph, Cognee, or custom

## Key difference from search

Search is **query-driven** (user asks, system retrieves). Memory would be **context-driven**: the proxy automatically injects relevant recalled context into LLM requests before forwarding them upstream.

jpmcb:
Fixes #81 

* ✨ Implements a new `pkg/memory` package and `memory.Driver`

---

**Continue Tasks:** ✅ 2 no changes — [View all](https://hub.continue.dev/inbox?pr=https%3A%2F%2Fgithub.com%2Fpapercomputeco%2Ftapes%2Fpull%2F86&utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/api/mcp/memory.go
+++ b/api/mcp/memory.go
@@ -1,0 +1,70 @@
+package mcp
+
+import (
+	"context"
+	"encoding/json"
+	"fmt"
+
+	"github.com/modelcontextprotocol/go-sdk/mcp"
+
+	"github.com/papercomputeco/tapes/pkg/memory"
+)
+
+var (
+	memoryRecallToolName    = "memory_recall"
+	memoryRecallDescription = "Recall facts from the tapes memory layer. Given a node hash (a position in the conversation DAG), returns extracted facts that are relevant to that position. Use this to retrieve persistent knowledge from past conversations."
+)
+
+// MemoryRecallInput represents the input arguments for the MCP memory_recall tool.
+type MemoryRecallInput struct {
+	Hash string `json:"hash" jsonschema:"the node hash identifying a position in the conversation DAG to recall facts for"`
+}
+
+// MemoryRecallOutput represents the structured output of a memory recall.
+type MemoryRecallOutput struct {
+	Facts []memory.Fact `json:"facts"`
+}
+
+// handleMemoryRecall processes a memory recall request via MCP.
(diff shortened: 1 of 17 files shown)


api/mcp/memory.go
```diff
@@ -0,0 +1,70 @@
...
+	if input.Hash == "" {
+		return &mcp.CallToolResult{
+			IsError: true,
+			Content: []mcp.Content{
+				&mcp.TextContent{Text: "hash is required"},
+			},
+		}, MemoryRecallOutput{}, nil
+	}
+
+	facts, err := s.config.MemoryDriver.Recall(ctx, input.Hash)
```
vkovac2: For query-based backends (e.g., Cognee), there is a possible mismatch here since they expect queries, not hashes. Consider resolving the hash via `DagLoader` to derive a query and pass that to the driver.

vkovac2 (commented):
Nice work! Left two line comments about a similar concern: cognee is query-based, not hash-based. Something else that crossed my mind is whether there will be one dataset per session (or per conversation), or a single shared dataset (for `Store`)

PR #118, opened 2026-02-17
feat ✨: implements `pkg/logger`

jpmcb:
* ✨ Implements `pkg/logger` for a thinnly wrapped `slog/logger` and zerolog alongside Charm's `log` package for pretty logging.

---

**Continue Tasks:** ✅ 3 no changes — [View all](https://hub.continue.dev/inbox/pr/papercomputeco/tapes/118?utm_source=github_pr&utm_medium=pr_body&utm_campaign=continue_tasks)

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -2,11 +2,11 @@ package api
 
 import (
 	"fmt"
+	"log/slog"
 	"net"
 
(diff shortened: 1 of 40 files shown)


bdougie (commented):
Two inline non blocking comments below — otherwise this looks great.

PR #141, opened 2026-03-04
🔧 fix: Parse OpenAI-format content in Ollama provider

mtfoley:
Resolves #137.

## Changes

- **Fix fallback detection** — The previous fallback triggered on `json.Unmarshal` errors, but Go's decoder silently zero-values type-mismatched fields (array→string) rather than erroring. Added `hasLostContent()` detection after successful unmarshal: if a message has a role but lost its content, images, and tool calls, fall back to the OpenAI parser.
- **Fix response fallback** — Detect OpenAI-format responses (which use a `choices` array instead of a top-level `message` field) by checking for zero-valued message role.
- **Extract OpenAI parsing** — Moved parsing logic into exported `ParseRequestPayload`/`ParseResponsePayload` functions so the Ollama provider can reuse them.
- **Handle OpenAI-format usage in SSE** — Extract `usage` object from Ollama SSE chunks that use OpenAI format (e.g. OpenCode).

## Test plan

- [x] OpenCode-style request with array content ([exact scenario from #137](https://github.com/papercomputeco/tapes/issues/137))
- [x] Native Ollama string content still works (no false positive)
- [x] OpenAI-format response with `choices` array and `usage` object
- [x] All existing Ollama tests pass (32/32)
- [x] All existing OpenAI tests pass (22/22)
- [x] All proxy tests pass (53/53)


bdougie:
Left a comment on the issue, but please reach out if you want to be an early tester of our new cli. Going to close this for now.

PR #176, opened 2026-04-06
 feat: support Codex ChatGPT OAuth login

BakedSoups:
###  This PR makes it so we can use the 
native login session for codex so we  can do:
```
  codex login
  tapes start codex
```
instead of only having the api as an option 

Implementing Codex needed a different flow because its ChatGPT login
  path is built around a local ChatGPT-style proxy surface, not the API flow used by OpenCode and Claude, so we added a Codex-specific local proxy path and raw Responses relay using `github.com/gofiber/contrib/v3/websocket` to capture and store turns.

  Tapes now:
  - reads ~/.codex/auth.json
  - uses the existing OAuth session
  - routes Codex traffic to chatgpt.com
  - handles the local proxy surface Codex expects
  - captures and stores turns through a Codex-specific raw Responses
    relay

  Test note:
  - updated pkg/credentials/codex_test.go to match current behavior:
    preserve OAuth tokens and add OPENAI_API_KEY


BakedSoups:
Where in Tapes do Claude/OpenCode responses get turned into a ‘turn’, and can we plug Codex websocket events into that same layer?

BakedSoups:
got it working with a python proxy for now : 
  What it does:
  - listens locally on 127.0.0.1:8765
  - accepts Codex POST /v1/responses
  - rewrites that to https://chatgpt.com/backend-api/codex/responses
  - forwards the original body with minimal header changes
  - reads the streamed response events
  - reconstructs one final turn from those events
  - optionally posts that reconstructed turn to Tapes ingest

  How the recording works:
  - when it sees response.created, it opens a turn
  - while the stream is active, it collects text from events like response.output_text.delta
  - when it sees response.done, response.completed, or response.failed, it closes the turn
  - it builds:
      - a minimal request: model + user message text
      - a minimal response: model + assistant text
  - then it logs turn_reconstructed
  -   if --ingest-url is set, it POSTs that to /v1/ingest

bdougie:
Closing this in favor of your other pr

PR #182, opened 2026-04-10
Feat/codex transparent proxy

BakedSoups:


bdougie:
Hey @BakedSoups we are moving these start features to a new unreleased cli called `paper` we will needed start this work over there when ready.

PR #172, opened 2026-03-26
🧹 chore: Remove tapes sync command and backfill package

bdougie:
## Summary
- Remove `tapes sync` CLI command and `pkg/backfill/` package — this functionality is moving to the [moonlander](https://github.com/papercomputeco/moonlander) project
- Remove sync-related telemetry events (`EventSyncPush`, `EventSyncPull`, `CaptureSyncPush`, `CaptureSyncPull`)
- `tapes skill sync` (skill file copying) is **not** affected

## Test plan
- [x] `go build ./...` compiles successfully
- [x] `go test ./pkg/telemetry/...` passes
- [ ] Verify no remaining references to removed packages

PR #184, opened 2026-04-15
🔧 fix: tapes sync ingests Claude subagent transcripts

bdougie:
## Summary

`tapes sync` was silently dropping Claude Code subagent (sidechain) turns. Claude moved those turns out of the parent `<session>.jsonl` and into a new `<session>/subagents/agent-*.jsonl` layout, and since those API calls never passed through the tapes proxy, backfill had no existing DB node to update — so the data landed nowhere.

This PR teaches `backfill.Run` a second pass: every transcript assistant entry that didn't match an existing node gets inserted directly into the store as a fresh `merkle.Node`. Node hashes are content-addressable, so re-running sync is a true no-op — nothing duplicates.

Subagent transcripts pull `agent_name` from the sibling `agent-<id>.meta.json`; if the meta isn't on disk yet, the entry is deferred to a future run rather than inserted with a fallback name (AgentName is part of the hash, so "claude" and "Explore" would produce two distinct rows for the same logical turn).

PR #179 already taught `/v1/sessions/summary` to tolerate dangling parents via `Truncated` / `MissingParent`, so these inserted nodes surface as their own truncated chains with no special-case handling needed.

## Scope

FYI @jpmcb. This is the **short-term unblock for today's demo** so Claude Code users (hi Tom) stop losing subagent data. The long-term home is the new `paper` CLI — see papercomputeco/paper#6, which ports moonlander's multi-agent engine (claude + codex + copilot + opencode) into `paper sync` and carries this same subagent fix forward. Once paper#6 ships, tapes#172 deletes this entire code path so we aren't running three copies of transcript ingestion.

## Smoke test on a real ~/.tapes copy

First sync on a fresh copy of my `~/.tapes/tapes.sqlite`:

> **5,169 new rows inserted** — 335 of them correctly tagged `agent_name=Explore` from subagent meta files. 29,968 pre-existing proxy-written rows had their token counts filled in. Zero errors.

Immediately re-running the same command:

> **14 new rows** (sessions Claude wrote between runs) plus **5,497 existing rows correctly deduped by hash**. This is exactly what idempotency looks like — the second-run "skipped" counter is the receipt, not a failure. If you ran sync a third time without touching Claude, you'd see 0 new rows and ~5,511 deduped.

DB grew 181,955 → 187,138 total nodes across the two runs.

## Test plan

- [x] `go test ./pkg/backfill/...` — all Ginkgo specs green (25 including the new subagent + idempotency tests)
- [x] `go test ./...` — full unit suite green
- [x] `golangci-lint run ./pkg/backfill/...` clean
- [x] `go build ./...` clean
- [x] Smoke test on a /tmp copy of a real `~/.tapes/tapes.sqlite` (numbers above)
- [ ] Reviewer spot-check on their own `~/.claude/projects` to confirm subagent rows (e.g. `agent_name=Explore`) show up in `tapes deck`


bdougie:
closing in favor of https://github.com/papercomputeco/paper/pull/6

bdougie:
Reopened and marked ready for review — this is the **short-term unblock** for today's demo. Claude Code users currently have a broken `tapes sync` because Claude moved subagent turns into `<session>/subagents/agent-*.jsonl` and the legacy walker drops them. This PR fixes that in place so nothing has to wait on the longer-term migration.

**Sequencing for the three PRs in flight:**

1. **This PR (tapes#184) — now, for demo.** Merges the claude subagent fix into `pkg/backfill`. Keeps `tapes sync` working until paper CLI ships.
2. **papercomputeco/paper#6 — later, when paper CLI is ready.** `paper sync [agent...]` is the real long-term home: multi-agent (claude + codex + copilot + opencode), ported from moonlander, carries forward the same subagent fix that's in this PR. Holding for paper readiness.
3. **papercomputeco/tapes#172 — after #6 lands.** The existing \"Remove tapes sync command and backfill package\" PR (already open, authored by @bdougie). Refresh and merge once users have `paper sync` in their hands, so nothing calls into the legacy path anymore.

tl;dr: land this for the demo, then paper#6, then tapes#172 to close the loop.

PR #185, opened 2026-04-15
✨ feat: tapes menu — menu bar app showing the tapes logo

bdougie:
## Summary

Screenshot 2026-04-16 at 11 50 44 AM

- Add macOS menu bar app (`tapes menu`) showing the tapes logo
- Auto-launch the menu from `tapes serve` (one process per workstation, guarded by `flock` on `~/.tapes/menu.pid`)
- Pre-render the icon at build time via `//go:generate` so the runtime menu binary does not link `oksvg`/`rasterx`
- No-op on non-macOS platforms via build tags

## Lifecycle

The menu is a persistent app, not a child process of `tapes serve`. Once spawned it runs independently and survives `tapes serve` restarts — re-running `tapes serve` is a no-op for the menu (the new process exits silently because the lock is held). The user quits the menu manually via the **Quit** item in the menu bar. The kernel auto-releases the `flock` on process exit, so a hard kill cannot strand the lock.

## Test plan

- [ ] Run `tapes serve` on macOS and verify the tapes icon appears in the menu bar
- [ ] Re-run `tapes serve` while the first menu is still up and verify no second icon appears (flock holds)
- [ ] Click **Quit** in the menu and verify the icon disappears and \`~/.tapes/menu.pid\` no longer holds the lock (a subsequent `tapes serve` spawns a new menu)
- [ ] \`Ctrl+C\` \`tapes serve\` while the menu is up and verify the menu **stays running** (this is intentional — the menu owns its own lifecycle)
- [ ] Build on Linux and confirm `tapes serve` does not attempt to spawn anything (build tags)


public/tapes-logo.svg
jpmcb, line 1: Why this SVG?

cmd/tapes/menu/menu.go
```diff
@@ -0,0 +1,139 @@
...
+	"github.com/getlantern/systray"
+	"github.com/spf13/cobra"
+	"github.com/srwiley/oksvg"
+	"github.com/srwiley/rasterx"
+
+	"github.com/papercomputeco/tapes/pkg/start"
+)
+
+//go:embed tapes-logo.svg
+var tapesLogoSVG []byte
```
jpmcb: Instead of embedding the SVG and having to call `renderTemplateIcon` each time, we should have a small `generate` step that can just populate a slice of bytes without having to bring the whole SVG and pre-processing with it.

pkg/start/manager.go
```diff
@@ -28,6 +28,7 @@ type AgentSession struct {
 type State struct {
 	Version          int            `json:"version"`
 	DaemonPID        int            `json:"daemon_pid"`
+	MenuPID          int            `json:"menu_pid,omitempty"`
```
jpmcb: The tapes menu sub-program should handle its own `.pid` file, probably in `~/.tapes/menu.pid` so we don't have to reference it in this manager state.

public/tapes-logo.svg
bdougie: My assumption is SVG is the best option for this, but I see png is also possible. https://bjango.com/articles/designingmenubarextras/#file-format

cmd/tapes/menu/menu.go
bdougie: Done in 27fe753. Added a small generator at `cmd/tapes/menu/internal/icongen` invoked via `//go:generate` — the menu now embeds the pre-rendered `tapes-logo.png` and oksvg/rasterx no longer participate in the runtime path. Re-run with `go generate ./cmd/tapes/menu/...` if the SVG changes.

pkg/start/manager.go
bdougie: Done in c7c3a81. Dropped `MenuPID` from `start.State` entirely; `menucmder.Spawn` now reads/writes `~/.tapes/menu.pid` itself (resolved via the dotdir manager). The pid file is removed when the menu process exits, so a stale entry can't suppress a future spawn. This also let me delete the `MenuPID` preservation hack in `tapes start`.

bdougie:
Self-review pass addressed two races and a coverage gap. Pushed `3903a6c`:

**Fixed**
- **PID reuse footgun:** the previous `signal(0)` check on `~/.tapes/menu.pid` could be defeated by macOS recycling a dead menu's PID onto an unrelated process. Replaced with `syscall.Flock` held by the menu binary for its lifetime — the kernel auto-releases on process death, so SIGKILL can't strand the lock.
- **TOCTOU between concurrent `tapes serve` invocations:** two parallel spawns could both pass the alive check and both fork; second's cleanup goroutine would later unlink the pid file while the second menu was still running, hiding it from future spawns. Moving enforcement into the menu binary itself eliminates the check-then-fork window — `Spawn` is now a plain `exec.Start`, and the new process exits silently if the lock is held.
- **Test coverage:** added `pidfile_test.go` with Ginkgo specs for holder/contender, release-on-close, stale-pid-file-is-ignored, and pid persistence.

**Kept by design**
The menu still spawns from `tapes serve` only — agent count comes from `start.State.Agents`, which `tapes start <agent>` writes. Running `tapes serve` standalone correctly shows "Running: 0" since no agents are launched.

public/tapes-logo.svg
jpmcb: No, I mean, why this svg committed to the repo when this also carries `cmd/tapes/menu/tapes-logo.png` and `cmd/tapes/menu/tapes-logo.svg`?

cmd/tapes/menu/menu.go
```diff
@@ -0,0 +1,108 @@
...
+				// Informational only — proceed even if we cannot persist
+				// the pid for human inspection.
+				_ = err
+			}
+
+			systray.Run(onReadyFor(configDir), onExit)
+			return nil
+		},
+	}
+}
```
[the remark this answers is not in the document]
bdougie: Pushing back on this one. `pollAgentCount` does not outlive the parent — `systray.Run` is the entire `RunE` body, so when the user clicks Quit and `systray.Quit()` returns, `onExit` fires, `RunE` returns, `main()` returns, and the process exits microseconds later. The ticker goroutine doesn't leak; it's reaped by process exit before it can tick again. Adding a done channel would be 6 lines of plumbing for no observable benefit, which conflicts with the project's "don't add code beyond what the task requires" guidance. Leaving as-is.

cmd/tapes/menu/spawn.go
```diff
@@ -0,0 +1,52 @@
...
+	cmd := exec.CommandContext(context.Background(), execPath, args...)
+	cmd.Stdout = nil
+	cmd.Stderr = nil
+
+	if err := cmd.Start(); err != nil {
+		log.Warn("could not start menu bar app", "error", err)
+		return
+	}
+
+	log.Info("spawned menu bar app candidate", "pid", cmd.Process.Pid)
```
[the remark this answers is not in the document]
bdougie: Fixed in e4dc3a5, but not via your suggestion — plain `exec.Command` triggers the `noctx` linter (which fired on this very file earlier in the PR). Switched to `exec.Command` with an explicit `//nolint:noctx` and a comment that calls out the deliberate "no cancellation" choice. The misleading "use Background context so it survives" comment is gone.

bdougie:
Re: Greptile's P1 lifecycle item — I went with option 2 (update PR description to match code), not option 1 (add SIGTERM cleanup). The current persistent-menu design is intentional per @bdougie's earlier directive to decouple the menu from `tapes start` and have it appear on server initialization, not be tied to it. Re-adding cleanup would reverse that.

Updated PR description and test plan now spell out:
- The menu is persistent and survives `tapes serve` restarts
- Re-running `tapes serve` is a no-op for the menu (flock is held)
- The user quits via the menu's Quit item; the kernel auto-releases the flock on process death
- Test plan now explicitly verifies the menu **stays running** after `Ctrl+C` on `tapes serve`

jpmcb (commented):
This doesn't look like it actually counts the number of tapes servers running (shows `Running: 0` on `tapes serve`)

Image

I would honestly think about removing the feature to count the number of tapes servers (for now) since that probably deserves its own API server that can handle state since relying on the `start` command manager: `mgr, err := start.NewManager(configDir)` won't work on remote tapes servers/apis.

public/tapes-logo.svg
bdougie: ah. that should be removed. Ideally we get a new one made from @isabensusan and it will be svg. We don't need both.

bdougie:
> This doesn't look like it actually counts the number of tapes servers running (shows `Running: 0` on `tapes serve`)
> 
Fair. I can remove, we can think more about that experience in a future iteration. cc / @isabensusan

public/tapes-logo.svg
bdougie: Actually upon further review. The SVG was provided from the assets I had assuming that was needed. Then claude code generated this icongen code to convert the SVG to a PNG which is what is actually needed. 

I've deleted the SVG and the icongen code that was adding more complexity.

yeazelm (approved):

bdougie:
closing in favor os `paper menu`

PR #188, opened 2026-04-16
🐛 fix: clamp deck --since to the 30d rolling window

bdougie:
## Summary

- `newDeckModel` previously trusted `filters.Since` as-is when positive, so a missing `--since` (0) or a value larger than 30d could leave the period label, cutoff, and displayed data disagreeing.
- Clamp `filters.Since` into `(0, maxSinceDuration]` before choosing the initial `timePeriod`, keeping the TUI on a coherent rolling 30-day window.
- Added Ginkgo specs covering zero, negative, over-sized, 7d, 24h, and sub-7d `--since` values.

## Test plan

- [x] `go test ./cmd/tapes/deck/ -run TestDeckCommander`

### DIFF
--- a/cmd/tapes/deck/tui.go
+++ b/cmd/tapes/deck/tui.go
@@ -166,17 +166,21 @@ func newDeckModel(query deck.Querier, filters deck.Filters, overview *deck.Overv
 		}
 	}
 
-	// Determine initial time period from filters
-	period := period30d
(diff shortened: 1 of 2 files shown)


bdougie:
Need to leverage the API moving forward. I will restart if the need arises, but for now working supporting future tools and goals.

PR #189, opened 2026-04-21
✨ feat: Introduce duckdb, refactor away other providers

jpmcb:
This PR is a large refactor that focuses tapes's storage and vector model on Postgres with `pg_duckdb` and `pgvector`. It also focuses direct database clients to exclusively use the API (i.e. `tapes deck`).

Closes PCC-366
Closes PCC-367

---

### Removes sqlite storage driver in favor of only Postgres

This removes sqlite in favor of centralizing on Postgres with the `pg_duckdb` extension for large recursive columnar queries. This allows us to dramatically simply the `storage.Driver` interface implementations and removes the need for a separate `merkle.DagLoader` interface: now, it's all just in the `storage.Driver` implementation for Postgres. We keep an `inmemory` storage driver implementation since that's very useful for tests but remove the ability to configure it during `serve`.

This also refactors away all the vector implementations in favor of `pgvector`.

This unlocks tapes to be more scalable, optimized for huge scale data, and deployable for our future cloud services.
 
### Introduces `tapes local` command for local development and use

The new `tapes local` allows for a user to run our postgres image and ollama on Docker - this is useful for super users and local development. `tapes local up` brings up the containers in Docker which can then be targeted with `tapes serve --postgres "postgres://tapes:tapes@localhost:5432/tapes?sslmode=disable"`

### Removes ent in favor of sqlc compiled sql queries

Since ent is mostly focused on generating down to entity related queries, we miss out on the recurse nature of duckdb queries and needing to inject `SET LOCAL duckdb.force_execution = true;` into the raw query context. So, instead, we use raw `.sql` queries with sqlc that compiles into the `pkg/storage/postgres/gensqlc/` package.

### Removes bespoke `pkg/storage/migrate/` in favor of `up` / `down` migrations with `go-migrate`

We had a hand rolled auto-migration package that could be brittle - this was removed in favor of using `go-migrate` and up down migrations in `migrations/...`. This also lets us get rid of `tapes migrate` since migrations are automatically managed by the tapes API when it connects to the Postgres db.

### Removes bespoke `tapesapi` and `tapesprox`: centralizes on one `tapes` binary

Removed the `Dockerfiles/tapesapi.Dockerfile` and seperate binary builds for `cli/tapesapi/...`. Everything goes through the monolithic `tapes` binary / build.

### `tapes deck` now uses the API without constant polling

The deck tui was doing near constant polling of essentially a databases entire dataset which would take forever to load. Now, it is better about not grabbing all the data from the db at once.

### Removes the tapes deck web view

This was some early experimentation on providing a web server locally for visualizing the tapes data. Since we're centralizing tapes on being more of a server technology ontop of the database, we're removing this in favor of future enhancements on a platform.

### `./cli/sqlite-import` utility for importing data into Postgres

After running `tapes local up`, this utility can be run to migrate a sqlite db into the local postgres.


pkg/storage/driver.go
```diff
@@ -4,20 +4,20 @@ package storage
...
 // The Driver is the primary interface for working with pkg/merkle - it handles
 // storage, retrieval, and traversal of nodes per the storage implementor.
-//
-// A Driver embeds merkle.DagLoader, so any storage.Driver is also a merkle.DagLoader.
-// This avoids the need for callers to cast a Driver to a DagLoader — they can pass
-// a Driver wherever a DagLoader is expected directly.
 type Driver interface {
-	// DagLoader provides read and traversal operations on the DAG.
-	// Get, GetByParent, and Ancestry come from this embedded interface.
-	merkle.DagLoader
```
jpmcb, line 20: This is a big change worth calling out: `storage.Driver` now just has the `DagLoader` interface methods directly: no more subverting one interface into another since everything should just be 1 driver now.

migrations/1776787153_ancestry_func.up.sql
```diff
@@ -0,0 +1,124 @@
...
+    total_duration_ns BIGINT,
+    prompt_duration_ns BIGINT,
+    project TEXT,
+    created_at TIMESTAMPTZ,
+    depth INTEGER,
+    has_usage BOOLEAN
+)
+LANGUAGE SQL
+AS $$
+    WITH RECURSIVE walk AS (
```
jpmcb, line 28: cc @yeazelm - this is one of the bigger bits to look at - this is the query you had written for CTEs walking the chains and is now re-written for `RECURSIVE` on duckdb's analytical engine:

https://duckdb.org/docs/current/sql/query_syntax/with#recursive-ctes

jpmcb:
@greptile-ai please review 🙏🏼

cmd/tapes/local/local.go
```diff
@@ -0,0 +1,358 @@
...
+		"--entrypoint", "sh",
+		"-v", postgresDir+":"+path.Dir(postgresDataPath),
+		image,
+		"-lc",
+		cmd,
+	); err != nil {
+		return fmt.Errorf("preparing postgres directory %q: %w", postgresDir, err)
+	}
+	return nil
+}
```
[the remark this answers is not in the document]
jpmcb: This is fine - the whole point is that the user would be using our image (same on the cloud platform).

jpmcb:
Force pushed to address Greptile's review -  the migration DSN now gets all the key/values so migrations don't silently fail. And defensively try both uint32 and uint64 so we don't silently loose points in the pg type conversions.

pkg/deck/http_query.go
```diff
@@ -20,19 +19,15 @@ import (
...
-	// floor: each /v1/sessions/summary call costs ~260ms regardless of
-	// page size on a real store, so smaller pages just multiply the
-	// total round-trip count for no per-row savings. 2000 is a safe
-	// page size for the bumped storage.MaxListLimit (5000).
-	httpQueryPageLimit  = 2000
-	httpQueryMaxSummary = 10000 // safety cap on total sessions paged through
+	// httpQueryOverviewLimit keeps the API-backed deck lightweight: fetch one
+	// recent page of rich session summaries, then render and drill into detail
+	// on demand. This avoids the old "page through the whole corpus" behavior.
+	httpQueryOverviewLimit = 25
```
yeazelm, line 25: I think this has an indirect effect of never fetching more data, so you are essentially making it a "just the 25 top hits" query. Since there isn't way to have the UI load more, you basically are hard capped on what you can see.

pkg/deck/http_query.go
```diff
@@ -84,22 +90,23 @@ type httpTurn struct {
...
+// smaller page returned below.
 func (q *HTTPQuery) Overview(ctx context.Context, filters Filters) (*Overview, error) {
-	all, err := q.fetchAllSummaries(ctx, filters)
+	page, err := q.fetchSummaryPage(ctx, "", filters, httpQueryOverviewLimit)
 	if err != nil {
 		return nil, err
 	}
 
+	all := page.Items
+
```
yeazelm: I think this doesn't paginate and pull more data anymore which leaves it as a single call now with the limit essentially being capped at 25.

cmd/tapes/serve/serve.go
```diff
@@ -58,20 +51,18 @@ type ServeCommander struct {
...
-	config.FlagEmbeddingTgt:    {Name: "embedding-target", ViperKey: "embedding.target", Description: "Embedding provider URL"},
-	config.FlagEmbeddingModel:  {Name: "embedding-model", ViperKey: "embedding.model", Description: "Embedding model name (e.g., nomic-embed-text)"},
-	config.FlagEmbeddingDims:   {Name: "embedding-dimensions", ViperKey: "embedding.dimensions", Description: "Embedding dimensionality"},
+	config.FlagProxyListen:    {Name: "proxy-listen", Shorthand: "p", ViperKey: "proxy.listen", Description: "Address for proxy to listen on"},
+	config.FlagAPIListen:      {Name: "api-listen", Shorthand: "a", ViperKey: "api.listen", Description: "Address for API server to listen on"},
+	config.FlagIngestListen:   {Name: "ingest-listen", Shorthand: "i", ViperKey: "ingest.listen", Description: "Address for ingest server to listen on (sidecar mode)"},
+	config.FlagUpstream:       {Name: "upstream", Shorthand: "u", ViperKey: "proxy.upstream", Description: "Upstream LLM provider URL"},
+	config.FlagProvider:       {Name: "provider", ViperKey: "proxy.provider", Description: "LLM provider type (anthropic, openai, ollama)"},
```
yeazelm, line 58: I tried this locally and for some reason they were swallowed, I set the provider as `anthropic` and upstream to the one needed for Claude Code and it worked, but these flags don't seem to make it to the right place anymore.

cmd/tapes/local/local.go
```diff
@@ -0,0 +1,358 @@
...
+	if err != nil {
+		return err
+	}
+	if err := prepareLocalPostgresDir(postgresDir, c.postgresImage); err != nil {
+		return err
+	}
+	if err := ensurePostgresContainer(c); err != nil {
+		return err
+	}
+	if err := ensureOllamaContainer(c); err != nil {
```
yeazelm, line 137: We probably also need to ensure the model that is needed for embeddings is pulled when doing this:

```
2026/04/24 14:49:44 WARN failed to generate embedding hash=97a3af412a4637cf070420435945ae1048bfb6066d42732274f12734e952fd61 error="embedding failed: ollama returned status 404: {\"error\":\"model \\"nomic-embed-text\\" not found, try pulling it first\"}"
```

jpmcb: Those models are quite large: we mount the user's `~/.ollama/` directory so normal "ollama pull ..." works and persists those models and we can just use the ones already there.

cmd/tapes/local/local.go
```diff
@@ -0,0 +1,358 @@
...
+}
+
+func dockerOutput(args ...string) (string, error) {
+	cmd := exec.CommandContext(context.Background(), "docker", args...)
+	out, err := cmd.CombinedOutput()
+	if err != nil {
+		return "", fmt.Errorf("docker %s: %w\n%s", strings.Join(args, " "), err, strings.TrimSpace(string(out)))
+	}
+	return string(out), nil
+}
```
yeazelm, line 358: These don't have cancellation, so a hung docker run will block `tapes local up` forever.

cmd/tapes/serve/serve.go
```diff
@@ -202,32 +167,25 @@ func NewServeCmd() *cobra.Command {
 func (c *ServeCommander) run() error {
 	c.logger = logger.New(logger.WithDebug(c.debug), logger.WithPretty(true))
 
-	// Create shared driver (satisfies storage.Driver, which embeds merkle.DagLoader)
-	driver, err := c.newStorageDriver()
+	driver, err := postgres.NewDriver(context.TODO(), c.postgresDSN)
```
yeazelm, line 170: We could probably replace the TODO with a `signal.NotifyContext` context so that Ctrl-C can cleanly close the database things.

cmd/tapes/local/local.go
jpmcb: We can smooth out the `tapes local` experience in the future: this PR's already ballooned quite abit. I'm not really worried about `tapes local` being the primary way people use or interface with tapes and honestly, `tapes local` will probably be mostly a development utility for the present.

cmd/tapes/serve/serve.go
jpmcb: I think that'd be a good future enhancement since I'd want to pipe a cancellation context down through the commander since we also have context that's on the telemetry client.

cmd/tapes/local/local.go
yeazelm: Yeah, I think this was the missing piece, the opinion on what embedding model to use shifted and I didn't have the local one there, so its more of a nice thing to pull the model. I think this was more of just a rough edge in the shift of preferred embedding model than this particular change.

bdougie:
This could be something addressed later, but I looked at `./build/tapes deck` and got an error because I was wasn't running the `serve`. For sure this is a local experience we can figure out, but it would be nice if the error is caught and the user is reminded to run `tapes serve` before running `tapes deck`. 

Screenshot 2026-04-24 at 3 25 19 PM

cmd/tapes/serve/serve.go
yeazelm: I just tried to reproduce to show the errors and they are gone. I'll go ahead and say this must have been a bad invocation I had in my shell history.

migrations/1776787153_ancestry_func.up.sql
yeazelm: Nice! Looks good!

yeazelm (approved):
`tapes deck` might be in a bad spot with the refactor but the rest looks goood!

jpmcb:
Pushed [`7fd4477`](https://github.com/papercomputeco/tapes/pull/189/commits/7fd4477c4193f91799fbb78a4ebda979c4907d1d) which addresses the tapes deck issues: still seeing some strange behavior related to some hanging leaves and Claude tool calling, but we're tracking that.

yeazelm:
Nice! I like the new pager! Feels like a reasonable trade off to let the TUI be responsive while letting the user get more data if they want.

PR #192, opened 2026-04-26
🔧 fix: Capture anthropic streaming with dedicated logic

yeazelm:
## Summary

Streaming responses from Anthropic Messages don't quite work as expected in non-local environments. The cloud sidecar reducer had three compounding bugs — body chunks overwriting instead of appending, dispatch firing on the first SSE frame instead of `EndOfStream`, and a `json.Marshal` error being swallowed into an empty POST body. The local proxy has its own working reducer but the two capture paths never shared code, so local would work and the cloud path didn't.

This PR fixes the regression and prevents its recurrence by consolidating streaming reduction into one library and tightening ingest visibility:

- **`pkg/capture` (new):** Shared streaming-reduction library. Anthropic is the first provider implemented; the state machine walks the full Messages wire format — text, `tool_use` with `input_json_delta` reassembly, extended thinking with signature, ping keep-alives, mid-stream `error` events, EOF before `message_stop`. 

- **`tapes-ingest` hygiene:** Every ingest failure used to collapse to HTTP 400. Splits responses into 400 / 422 / 502 via sentinel errors so we can distinguish envelope errors from unprocessable turns from downstream outages. Adds a Prometheus metric set on `/metrics`, a `GET /v1/ingest/nodes?agent=` verification surface for e2e tests, and a `contract_test.go` `DescribeTable` enumerating the full status-code × shape matrix.

- **Local proxy:** Routes Anthropic SSE through `capture.Reduce`. This allows both local and cloud paths to use the same code.

TL;DR is we now look for the full stream before trying to create the turn so that it will successfully parse into the data structure we expect. Since different providers will likely have different logic/implementations, this breaks the code out into a `capture` package that can be used for each provider.

## Testing

- [x] `GOEXPERIMENT=jsonv2 go test -race -count=1 ./pkg/capture/...` (24 specs: state machine across text / tool_use / thinking / mid-stream error / truncated stream, registry thread safety, canonical equivalence across 3 fixture pairs)
- [x] `GOEXPERIMENT=jsonv2 go test -race -count=1 ./ingest/...` (status-code matrix `DescribeTable`, Prometheus scrape assertions, `GET /v1/ingest/nodes?agent=` filter + empty-result tests)
- [x] `GOEXPERIMENT=jsonv2 go test -race -count=1 ./proxy/...` (Anthropic-via-`capture` end-to-end, split-TCP-write regression for the legacy SSE reader, existing OpenAI/Ollama suites unchanged)
- [x] `GOEXPERIMENT=jsonv2 go build ./...` clean


pkg/capture/anthropic.go
```diff
@@ -0,0 +1,210 @@
...
+
+// NewAnthropicReducer returns an Anthropic reducer. The value is stateless
+// at the package level; per-turn state lives inside Reduce.
+func NewAnthropicReducer() Reducer {
+	return &anthropicReducer{}
+}
+
+func init() {
+	Register(ProviderAnthropic, NewAnthropicReducer())
+}
```
jpmcb: We shouldn't use `init`: when debugging or tracking down issues, it's _really_ hard to track what's happening when a "magic" init fires from a package and you end up needing global async state mechanisms (like the `registry`) to track that. `init` also can make testing difficult since it runs as soon as a package is imported.

We should simplify and just expect the caller to handle it in their call loops.

This would also eliminate the need for a package scoped registry of reducers with a RW mutex. The proxy or other consumers could just handle it like:

```go
anthReducer := capture.NewAnthropicReducer()
openAIReducer := capture.NewOpenAIReducer()

// etc. etc. - inside loop of handling SSE messages
authReducer.Reduce()
openAIReducer.Reduce()
```

or if we needed to be fancy, could just have it managed in a slice of reducers that the SSE handler manages itself:

```go
// during proxy bootstrapping ...
reducers := []capture.Reducer{}
reducers = append(reducers, capture.NewAnthropicReducer())
reducers = append(reducers, capture.NewOpenAIReducer())

// etc. etc. - inside loop of handling SSE messages
for _, r := range reducers {
    r.Reduce()
}
```

The only case where `init` can be really useful is if we provide some sort of "package" mechanism for people to bring their own reducer that automatically gets init'd / imported ... but I can't imagine we'll need that since there are a fairly limited number of LLM providers and most have aligned with OpenAI's API shape.

pkg/capture/reducer.go
```diff
@@ -0,0 +1,96 @@
...
+
+// registry holds the process-wide provider-to-reducer map. Each provider
+// package calls Register() from its init() to install itself. The mutex
+// guards against concurrent Register/Get from tests and late providers that
+// register outside init(); production init() is already serialized but the
+// lock keeps the invariant enforceable rather than convention-only.
+var (
+	registryMu sync.RWMutex
+	registry   = map[string]Reducer{}
+)
```
jpmcb: I also wouldn't recommend a package scoped mutex or registry:

jpmcb (commented):
Very nice work on this: tricky stuff. Just one architecture comment on expecting `init` and the global `registry` state.

pkg/capture/anthropic.go
yeazelm: This is great! I've pushed an update to refactor to avoid the `init` path.

jpmcb (approved):
Great work!! 👏🏼👏🏼

PR #194, opened 2026-04-27
🔧 fix: Actually pull images during "tapes local"

jpmcb:
Actually pull the images during `tapes local`. Also uses the `cliui` package for nicer output.


jpmcb:
Much nicer output:

image

yeazelm (approved):
Very nice!

PR #198, opened 2026-05-05
✨ feat: Wire ingest queue-depth gauge and add API server /metrics

yeazelm:
## Summary

Two related but independent instrumentation gaps closed in one PR (two commits). Both sit behind the cluster's `annotationAutodiscovery` configuration once a tapes release containing them lands and the per-tenant ingest pod annotation for the cloud.

### Commit 1 — Wire ingest worker queue depth gauge

The gauge `tapes_ingest_worker_queue_depth` was already registered in `ingest/metrics.go` but never `Set` anywhere — it shipped as flatline-zero. This commit:

- Adds `Pool.Len() int` to `proxy/worker/pool.go` (returns `len(p.queue)`).
- Calls `s.metrics.SetQueueDepth(s.workerPool.Len())` on both enqueue paths in `processTurn`.
- Tightens the gauge's Help text to flag that values reflect ingest-path observations only — the underlying `Pool` is shared with three proxy `Enqueue` call sites that don't currently update the gauge. Wiring those is filed as a follow-up.
- Adds a `Len()` test that exercises non-zero queue state via a stub-channel `Pool` (bypasses `NewPool` so workers don't drain before the assertion). The pre-existing drained-pool spec would have passed against a `Len() int { return 0 }` regression; this one wouldn't.

### Commit 2 — Expose Prometheus RED metrics from the API server

The tapes API server previously had no `/metrics`. 

- New `tapes_apiserver_requests_total{route,method,status}` counter, `tapes_apiserver_request_duration_seconds{route,method}` histogram, `tapes_apiserver_inflight_requests` gauge.
- `/metrics` mounted via `gofiber/adaptor/v2` (the Fiber-blessed shim — explicitly NOT `valyala/fasthttp/fasthttpadaptor`, since that mixes web frameworks).
- Fiber RED middleware ordered BEFORE `recover.New()` so panicked requests are still counted (a panic propagating through `c.Next()` would otherwise skip the post-Next counter increments and disappear from RED metrics).
- `resolveStatus` helper uses `errors.As` so a wrapped `*fiber.Error` (e.g. `fmt.Errorf("...: %w", fiber.ErrBadRequest)`) gets the wrapped status code in metrics, matching what Fiber's own ErrorHandler does.
- New `isRouterNotFound(err)` helper for the unmatched-route sentinel

Route labels stay templated (`/v1/sessions/:hash`, never the materialized URL) — bounded cardinality regardless of session count.

## Tests

- New `newTestApp()` helper for specs that exercise middleware behavior in isolation. Used by the panic, wrapped-error, and unmatched-attribution specs so they don't pollute the suite-level Server with `/_test/*` paths.
- Defensive contract test: registers a real `/` handler AND fires an unmatched URL, asserts the unmatched bucket stays distinct from the `/` row. Intentionally fragile — if anyone breaks `isRouterNotFound` or trips a related collision, this fails. Comment explicitly tells the future reader to read the rationale before papering over the assertion.
- `tapes_apiserver_*` cardinality contract verified: templated `:hash` collapses two materialized URLs onto one row; three never-registered URLs collapse to one `unmatched` row.

## Verified

- `go vet ./...` clean
- `make format` clean
- 44/44 specs in `api/`, 13/13 in `proxy/worker/`, ingest suite green
- Pre-PR review confirmed: `Pool.Len()` exported surface justified (cross-package use); `errors.As` usage textbook; constructor pattern matches `NewExampleStruct` paradigm per `AGENTS.md`; no new module deps; Ginkgo idioms consistent with neighboring suites.

### DIFF
--- a/api/metrics.go
+++ b/api/metrics.go
@@ -1,0 +1,163 @@
+package api
+
+import (
+	"errors"
+	"strconv"
+	"time"
+
+	"github.com/gofiber/adaptor/v2"
+	"github.com/gofiber/fiber/v2"
+	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
+)
+
+// Metrics is the Prometheus surface for the Tapes API server. Each Server
+// owns its own registry so tests can scrape in isolation; the production
+// path mounts /metrics on the Fiber app via NewServer.
+type Metrics struct {
+	registry *prometheus.Registry
+
+	requests *prometheus.CounterVec
+	duration *prometheus.HistogramVec
+	inflight prometheus.Gauge
+
+	// scrapeHandler is the cached /metrics Fiber handler. promhttp.HandlerFor
+	// with a non-nil Registry calls MustRegister(...) on the registry to
+	// instrument the scrape handler itself (promhttp_metric_handler_*),
+	// so building a fresh handler twice against the same Metrics would
+	// panic on the second call. Build once at construction; Handler()
+	// just returns this field.
+	scrapeHandler fiber.Handler
+}
+
+// NewMetrics constructs the Tapes API server's RED metrics. Labels stay
+// templated (`route`) rather than per-URL so :hash path params don't blow up
+// cardinality.
+func NewMetrics() *Metrics {
+	reg := prometheus.NewRegistry()
+	m := &Metrics{
+		registry: reg,
+		requests: prometheus.NewCounterVec(
+			prometheus.CounterOpts{
+				Name: "tapes_apiserver_requests_total",
+				Help: "Total HTTP requests handled by the Tapes API server, by route template, method, and status.",
+			},
+			[]string{"route", "method", "status"},
+		),
+		duration: prometheus.NewHistogramVec(
+			prometheus.HistogramOpts{
+				Name:    "tapes_apiserver_request_duration_seconds",
+				Help:    "Latency of requests handled by the Tapes API server.",
+				Buckets: prometheus.DefBuckets,
+			},
+			[]string{"route", "method"},
+		),
+		inflight: prometheus.NewGauge(prometheus.GaugeOpts{
+			Name: "tapes_apiserver_inflight_requests",
+			Help: "In-flight requests currently being handled by the Tapes API server.",
+		}),
+	}
+	reg.MustRegister(m.requests, m.duration, m.inflight)
+	m.scrapeHandler = adaptor.HTTPHandler(promhttp.HandlerFor(reg, promhttp.HandlerOpts{Registry: reg}))
+	return m
+}
+
+// Registry exposes the *prometheus.Registry so tests can scrape against the
+// same registry the middleware writes to.
+func (m *Metrics) Registry() *prometheus.Registry { return m.registry }
+
+// Middleware returns a Fiber handler that records request count + duration
+// per (route template, method, status). Templates like /v1/sessions/:hash
+// stay as the label value so the :hash path param never expands cardinality.
+//
+// Register this OUTSIDE recover.New() (i.e. via app.Use before recover) —
+// see resolveStatus for why.
+func (m *Metrics) Middleware() fiber.Handler {
+	return func(c *fiber.Ctx) error {
+		m.inflight.Inc()
+		defer m.inflight.Dec()
+
+		start := time.Now()
+		err := c.Next()
+
+		// Route().Path is the registered template (e.g. /v1/sessions/:hash);
+		// we fall back to a sentinel when Fiber's router never advanced past
+		// the middleware's own "/" entry, so unknown URLs don't each become
+		// their own series. Fiber v2 keeps c.route pointing at the last
+		// matched route — which for an unhandled request is this very
+		// middleware (path "/") — rather than nilling it, so we can't rely
+		// on an empty path. Detecting an unmatched request via the
+		// framework-emitted 404 fiber.Error is the contract that survives
+		// across Fiber versions: a real handler returning fiber.ErrNotFound
+		// is rare and harmless to bucket with router 404s here.
+		route := c.Route().Path
+		if route == "" || isRouterNotFound(err) {
+			route = "unmatched"
+		}
+
+		m.requests.WithLabelValues(route, c.Method(), strconv.Itoa(resolveStatus(c, err))).Inc()
+		m.duration.WithLabelValues(route, c.Method()).Observe(time.Since(start).Seconds())
+
+		return err
+	}
+}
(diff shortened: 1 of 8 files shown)


api/metrics.go
```diff
@@ -0,0 +1,151 @@
...
+}
+
+// Handler returns a Fiber handler that serves Prometheus text exposition
+// from this Metrics instance's registry. Mount it at /metrics with no
+// auth. Uses gofiber/adaptor (the Fiber-blessed shim over fasthttp) so
+// we don't introduce a direct fasthttp dep — Fiber already pulls
+// fasthttp in transitively.
+func (m *Metrics) Handler() fiber.Handler {
+	return adaptor.HTTPHandler(promhttp.HandlerFor(m.registry, promhttp.HandlerOpts{Registry: m.registry}))
+}
```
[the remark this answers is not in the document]
yeazelm: Done in `7df8b0c`. The handler is now built once at `NewMetrics()` time and cached in `Metrics.scrapeHandler`; `Handler()` just returns it. Added a regression spec (`Handler() can be called repeatedly without panicking`) so a future regression here surfaces in CI.

api/metrics.go
```diff
@@ -0,0 +1,151 @@
...
+// fragility — a handler that wants to surface a not-found resource almost
+// always uses a 4xx other than the framework's catchall, so collapsing both
+// into the sentinel row is fine in practice.
+func isRouterNotFound(err error) bool {
+	var e *fiber.Error
+	if !errors.As(err, &e) {
+		return false
+	}
+	return e.Code == fiber.StatusNotFound
+}
```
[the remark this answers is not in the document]
yeazelm: Good catch — fixed in `7df8b0c`. `isRouterNotFound` now uses `errors.Is(err, fiber.ErrNotFound)` to distinguish:

- Router-emitted catchall: a freshly-allocated `*fiber.Error` with message `"Cannot METHOD /path"` — does NOT match the package singleton, so `errors.Is(...)` returns false → still buckets as `unmatched`.
- Handler-emitted `fiber.ErrNotFound` (or wrapped via `fmt.Errorf("...: %w", fiber.ErrNotFound)`): matches the singleton via `errors.Is` → buckets on its registered route template.

Two new regression specs in `metrics_test.go` cover both raw and wrapped variants, so the misclassification you flagged would now surface in CI.

The tradeoff: a handler that returns `fiber.NewError(404, "custom message")` (custom 404 instance, not the singleton) would still bucket as `unmatched` since it doesn't match `errors.Is`. Documented in the updated `isRouterNotFound` comment.

jpmcb (approved):
👏 👏

jpmcb:
Very nice. Took this for a quick spin with `./build/darwin/arm64/tapes local up` and `./build/darwin/arm64/tapes serve --postgres "postgres://tapes:tapes@localhost:5432/tapes?sslmode=disable"`

```
❯ curl localhost:8081/metrics
# HELP promhttp_metric_handler_errors_total Total number of internal errors encountered by the promhttp metric handler.
# TYPE promhttp_metric_handler_errors_total counter
promhttp_metric_handler_errors_total{cause="encoding"} 0
promhttp_metric_handler_errors_total{cause="gathering"} 0
# HELP tapes_apiserver_inflight_requests In-flight requests currently being handled by the Tapes API server.
# TYPE tapes_apiserver_inflight_requests gauge
tapes_apiserver_inflight_requests 1
```

and the ingest server:

```
❯ curl localhost:8082/metrics
# HELP promhttp_metric_handler_errors_total Total number of internal errors encountered by the promhttp metric handler.
# TYPE promhttp_metric_handler_errors_total counter
promhttp_metric_handler_errors_total{cause="encoding"} 0
promhttp_metric_handler_errors_total{cause="gathering"} 0
# HELP tapes_ingest_worker_queue_depth Best-effort snapshot of pending items in the worker queue, as observed on the ingest enqueue path. The underlying Pool is shared with the proxy enqueue paths, which do not currently update this gauge — so the value reflects ingest-side observations only.
# TYPE tapes_ingest_worker_queue_depth gauge
tapes_ingest_worker_queue_depth 0
```

proxy/worker/pool.go
```diff
@@ -119,6 +119,14 @@ func (p *Pool) Enqueue(job Job) bool {
 	}
 }
 
+// Len returns the current number of jobs buffered in the queue. It is a
+// best-effort snapshot — workers may pick up items between the read and any
+// downstream observation — and is intended for metric instrumentation rather
+// than for routing decisions.
+func (p *Pool) Len() int {
+	return len(p.queue)
```
jpmcb, line 127: I wonder if there's a smarter way for us to communicate worker pool saturation since `p.queue` is a buffered channel - we can grab the channels capacity with `cap(p.queue)` and the "saturation" would effectively just be something like:

```go
func (p *Pool) Sat() float64 {
	return len(p.queue) / cap(p.queue)
}
```

ingest/metrics.go
```diff
@@ -44,7 +44,7 @@ func NewMetrics() *Metrics {
 		),
 		queueDepth: prometheus.NewGauge(prometheus.GaugeOpts{
 			Name: "tapes_ingest_worker_queue_depth",
-			Help: "Best-effort snapshot of pending items in the ingest worker queue.",
+			Help: "Best-effort snapshot of pending items in the worker queue, as observed on the ingest enqueue path. The underlying Pool is shared with the proxy enqueue paths, which do not currently update this gauge — so the value reflects ingest-side observations only.",
```
jpmcb, line 47: Since this is all in the same `ingest` package, we _could_ implement a custom [`prometheus.Collector`](https://github.com/prometheus/client_golang/blob/f6951953d56f596c99974f1fe46c2f8a393d9b73/prometheus/collector.go#L27) that handles computing the metric _"at scrape time"_, grabbing the metric directly in the `Collect` implementation - `len()` on the queue chan is thread safe so no reason we can't just real-time compute that value during a scrape:

```go
func (c *customIngestCollector) Collect(ch chan<- prometheus.Metric) {
    depth := float64(len(c.pool.queue))
    ch <- prometheus.MustNewConstMetric(c.depthDesc, prometheus.GaugeValue, depth)
}
```

Don't block on this - would be an interesting followup if we think it'd work!

proxy/worker/pool.go
yeazelm: We probably want to lean on emitting the two values and letting the ratio be derived in PromQL but this a good thought. We can follow up with adding in the ingest worker queue capacity to make this work in another PR.

ingest/metrics.go
yeazelm: I'll try it and do a follow up!

PR #211, opened 2026-05-21
🔧 fix: (merkle) keep one chain across mid-session harness drift (PCC-562)

jasonwc:
## Summary

[PCC-562](https://linear.app/paper-compute-co/issue/PCC-562). A single Claude Code conversation was fracturing into multiple disconnected roots in tapes because the proxy's content hash kept including parts of the bucket that drift between turns of the same conversation. Two commits, two families of drift; both addressed by projecting drift-prone fields out of the chain hash without touching what's stored.

Related to PCC-562

### Commit 1 — system-reminder + model drift

Claude Code prepends rotating `<system-reminder>` blocks (date, skills, MCP inventory) to every user turn after the first, and routes one conversation across multiple models (haiku pre-flight, then opus). Both changed the hashed content of "the same" prior turn, so the chain couldn't link and one conversation fractured into multiple roots.

**Fix:** strip harness-tag spans (`<system-reminder>`, `<command-*>`, `<local-command-*>`), normalize whitespace in the surviving prose, and drop `Model`/`Provider`/`AgentName` from the chain hash.

### Commit 2 — tool_use defaults + thinking signatures

Same fracture pattern, two more surfaces. The streamed assistant turn captures optional zero defaults (e.g. Edit's `"replace_all": false`) that the harness drops on re-send, and Anthropic's `thinking_signature` is emitted on the live stream but omitted on re-send. Both forked the chain mid-conversation.

**Fix:** prune zero-valued keys from `tool_use.ToolInput` (recursive for nested maps) and drop `ThinkingSignature` from the hash.

### Projection strategy

The raw `Bucket` is still stored verbatim — display, labels, search, and audit see exactly what the model received. Only `computeHash` (`pkg/merkle/node.go`) runs the bucket through `ProjectContent` (`pkg/merkle/projection.go`) and a `hashableBucket` subset that strips drift-prone, non-user-authored fields before SHA-256. Same content (post-projection) → same hash → chain links.

### Result

Same captured Claude Code session, reproduced across iterations of the fix:

| Stage | Nodes | Roots | Longest chain | Branch points | Leaves |
|---|---|---|---|---|---|
| Before (PCC-562 evidence)            | 51 | 3 | 42 | — | — |
| After commit 1                       | 53 | 2 | 44 | 5 | 8 |
| After commit 2                       | 49 | 2 | 44 | 2 | 4 |

The two remaining roots are: the conversation itself, and an unrelated `quota` count_tokens ping the harness fires as a separate API call. The two remaining branch points are legitimate (haiku topic-summary vs the real opus response; two genuinely different user inputs at one point in the chain) — not drift-driven sibling duplicates.

## Test plan

- [ ] `go test ./pkg/merkle/...` passes
- [ ] `go test ./pkg/capture/... ./pkg/sessions/... ./pkg/backfill/... ./pkg/storage/inmemory/... ./pkg/publisher/... ./proxy/...` passes
- [ ] Local Postgres truncate → run a Claude Code session → verify one connected root for the conversation, one root for the quota ping
- [ ] Walk `parent_hash` from the deepest leaf to the original `"Hey Claude!"` user turn without breaking

### DIFF
--- a/pkg/merkle/node.go
+++ b/pkg/merkle/node.go
@@ -78,13 +78,44 @@ func (n *Node) computeHash() string {
 		parent = *n.ParentHash
 	}
 
-	// Marshal to JSON using an inline struct for hash computation
+	// Marshal to JSON using an inline struct for hash computation.
+	//
+	// The hash is computed over a projection of the bucket, not the
+	// bucket itself, so that "the same logical conversation turn"
+	// hashes the same across the kinds of drift agent harnesses
+	// introduce between requests in a single conversation. Concretely:
+	//
+	//   * Content text inside <system-reminder>, <command-*>, and
+	//     <local-command-*> spans is stripped (the harness rotates
+	//     this preamble — clock ticks, skills load, MCP inventory
+	//     shifts).
+	//   * Blank-line whitespace drift in the surviving prose is
+	//     normalized (the harness re-serializes user text and
+	//     occasionally inserts a stray newline).
+	//   * Routing metadata — Model, Provider, AgentName — is dropped
+	//     entirely. A single conversation may legitimately fan a
+	//     pre-flight call to Haiku and the main work to Opus; that is
+	//     a routing decision, not a new turn.
+	//
+	// The raw n.Bucket is left untouched so storage / display / labels
+	// / search still see exactly what the model received, including
+	// model identity. See PCC-562.
+	type hashableBucket struct {
+		Type    string             `json:"type"`
+		Role    string             `json:"role"`
+		Content []llm.ContentBlock `json:"content"`
+	}
+
 	data, err := json.Marshal(struct {
-		Parent  string `json:"parent"`
-		Content Bucket `json:"content"`
+		Parent  string         `json:"parent"`
+		Content hashableBucket `json:"content"`
 	}{
-		Parent:  parent,
-		Content: n.Bucket,
+		Parent: parent,
+		Content: hashableBucket{
+			Type:    n.Bucket.Type,
+			Role:    n.Bucket.Role,
+			Content: ProjectContent(n.Bucket.Content),
+		},
 	})
 	if err != nil {
 		panic("failed to marshal hash input: " + err.Error())
--- a/pkg/merkle/node.go
+++ b/pkg/merkle/node.go
@@ -92,6 +92,11 @@ func (n *Node) computeHash() string {
 	//   * Blank-line whitespace drift in the surviving prose is
 	//     normalized (the harness re-serializes user text and
 	//     occasionally inserts a stray newline).
+	//   * Zero-valued keys are pruned from tool_use ToolInput so the
+	//     streamed capture (which has Edit's "replace_all": false)
+	//     dedups with the re-sent history (which omits the default).
+	//   * ThinkingSignature is dropped — present on the live stream,
+	//     absent when the harness re-sends the same thinking block.
 	//   * Routing metadata — Model, Provider, AgentName — is dropped
 	//     entirely. A single conversation may legitimately fan a
 	//     pre-flight call to Haiku and the main work to Opus; that is
--- a/pkg/merkle/projection.go
+++ b/pkg/merkle/projection.go
@@ -143,16 +143,16 @@ func stripTaggedSpan(text, tag string) string {
 			return text
 		}
 		rest := text[start:]
-		end := strings.Index(rest, closeTag)
-		if end == -1 {
+		_, after, ok := strings.Cut(rest, closeTag)
+		if !ok {
 			return text[:start]
 		}
-		text = text[:start] + rest[end+len(closeTag):]
+		text = text[:start] + after
 	}
 }
 
 var (
-	trailingLineSpace = regexp.MustCompile(`[ \t]+\n`)
+	trailingLineSpace   = regexp.MustCompile(`[ \t]+\n`)
 	consecutiveNewlines = regexp.MustCompile(`\n{2,}`)
 )
 
--- a/pkg/merkle/projection.go
+++ b/pkg/merkle/projection.go
@@ -65,7 +65,9 @@ func ProjectContent(blocks []llm.ContentBlock) []llm.ContentBlock {
 		if len(b.ToolInput) > 0 {
 			b.ToolInput = pruneZeroValues(b.ToolInput)
 		}
-		b.ThinkingSignature = ""
+		if b.ThinkingSignature != "" {
+			b.ThinkingSignature = ""
+		}
 		out = append(out, b)
 	}
 	return out
--- a/pkg/merkle/projection.go
+++ b/pkg/merkle/projection.go
@@ -39,6 +39,15 @@ var HarnessTags = []string{
 //     the surviving prose cannot fork the chain (see PCC-562's
 //     "57a58 >" case).
 //  3. Drops any block that became empty after stripping.
+//  4. For tool_use blocks, drops zero-valued keys from ToolInput. The
+//     streamed assistant capture sees every key the model emitted
+//     including optional defaults like Edit's "replace_all": false;
+//     when that same turn is re-sent as history the harness strips
+//     those defaults. Without this step the two captures hash to
+//     different nodes and the chain branches.
+//  5. For thinking blocks, drops ThinkingSignature. Anthropic emits
+//     the signature on the live stream but the harness omits it on
+//     re-send; same shape as (4).
 //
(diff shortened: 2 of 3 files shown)


jasonwc:
Remaining branches after this PR — not drift, but worth a follow-up

  Inspecting the post-fix trace, the conversation root has two surviving siblings at depth 1.
   Both are real (different content) but both are also harness-internal — the user never
  wrote either prompt and never sees either response.

  1. Haiku summary call (one per session)

  Claude Code fires a single call to claude-haiku-4-5 very early in the session, re-routing
  the first user turn to a cheaper model with a different system prompt. The response is a
  structured JSON title — e.g. {"title": "Run exercise-claude-harness-basic skill"} — used to
   label the session in Claude Code's UI.

  - Why it shows as a fork: we dropped Model from the chain hash, so the user side of the
  haiku call dedups with the user side of the opus call (both children of the same root). The
   haiku and opus responses are genuinely different content, so they correctly become
  siblings.
  - Volume: once per session.
  - Telemetry value: real — cost, tokens, model used. Hiding it means we can't answer "how
  much do we spend on session titles?"

  2. Suggestion mode (potentially many per session)

  When the user goes idle, Claude Code fires a side-channel request whose user message starts
   with the literal prefix [SUGGESTION MODE: Suggest what the user might naturally type next
  into Claude Code.]. The model returns a short predicted phrase. The user never types the
  prompt and never sees the response — Claude Code uses it for autocomplete-style hints in
  the input bar.

  - Why it shows as a fork: it's a real second user-role turn from the same parent, with
  content different from whatever the user actually typed. We can't make them hash-equal
  (genuinely different content), so the only way to keep one chain is to classify and route
  the side-channel.
  - Volume: fires on every idle tick. In a long session, easily many leaves.
  - Telemetry value: low. The prompt is a fixed harness template; the response is throwaway.

  Thoughts on this PR

  This PR fixes chain fracture from drift — the hashed bucket changing across captures of the
   same logical turn. The cases above are a different shape: fracture from genuine sibling
  branches caused by harness-internal side-channels, where there isn't drift to project away.

  Suggesting we merge this PR as-is and address the side-channels in a follow-up:

  - Suggestion mode is worth doing soon — high volume, no user value, single-prefix detection
   at the proxy. Cheapest version: drop the request from storage. Heavier alternative: store
  with a flag, filter at read time (preserves audit/cost telemetry).
  - Haiku summary is lower priority — once per session, real cost telemetry value, rendering
  layer (deck-style label grouping) already collapses it visually. Defensible to leave.
  - General shape of the follow-up: a single "is this an agent-internal side-channel?"
  classifier at the proxy that recognizes known patterns. One place to extend as more harness
   internals surface.

  Scoping this out of PCC-562 keeps the diff narrow and the hashing fix reviewable on its
  own.

pkg/merkle/node.go
```diff
@@ -78,13 +78,49 @@ func (n *Node) computeHash() string {
...
+	//     occasionally inserts a stray newline).
+	//   * Zero-valued keys are pruned from tool_use ToolInput so the
+	//     streamed capture (which has Edit's "replace_all": false)
+	//     dedups with the re-sent history (which omits the default).
+	//   * ThinkingSignature is dropped — present on the live stream,
+	//     absent when the harness re-sends the same thinking block.
+	//   * Routing metadata — Model, Provider, AgentName — is dropped
+	//     entirely. A single conversation may legitimately fan a
+	//     pre-flight call to Haiku and the main work to Opus; that is
+	//     a routing decision, not a new turn.
```
jpmcb, line 103: This makes me think we should just hoist `ProjectContent(...)` directly into `Bucket`, which is intended to be the "unit" for content-addressing, and hoist any incompatible data from `Bucket` into `Node`. Those structs probably then become:

```Go
type Bucket struct {
  Type string `json:"type"`
  Role string `json:"role"`
  ProjectedContent []llm.ContentBlock `json:"content"`
}

type Node struct {
  // Hoisted up metadata from Bucket that may change on the fly

  // Model may change between subagent calls to Haiku etc.
  Model string `json:"model"`

  // I suppose technically Provider could change if someone was on an open weight model
  // and shifted to a different inference provider of that model
  // It makes much more sense as metadata anyways.
  Provider string `json:"provider"`

  // If we end up supporting rehydration of generic sessions from one agent harness
  // to another, this will break things pretty quickly.
  // It's also "just metadata" and should probably be hoisted up.
  AgentName string `json:"agent_name,omitempty"`

  Hash ...
  ParentHash ...
  Bucket ...
  // etc. etc. other DAG node metadata
}
```

This way, we don't end up with this drift from

```
Node --> Bucket --> hashableBucket
```

Note: this would be a bigger refactor since `merkle.Bucket` gets populated in several places.

jpmcb (commented):
Very nice discovery work here, makes sense that some of this metadata was breaking the `Bucket --> Node` content-addressing bits - just one comment on hoisting up the inline struct to `Bucket`

pkg/merkle/node.go
yeazelm: I think I like this but I'm a bit concerned about conflating two things and merging them accidentally:

* I think we need something to normalize the turn to make it less likely to have different hashes for what are essentially the same turn. 
* But I also think we need to make sure we store all the data as it is sent to the model as well. The model sees all the data, not our normalized view, so differences in behavior or just plain different tool call data is still tracked. 

Maybe I'm conflating them and this is fine, but I'd like to hear folks thoughts on this?

jasonwc: No I think thats right. As designed currently, the existing bucket hold **all** the data and the hashable bucket holds the projection. If we just project into bucket, I think that results in loss.

pkg/merkle/projection.go
```diff
@@ -0,0 +1,169 @@
...
+			if projected == "" {
+				continue
+			}
+			b.Text = projected
+		}
+		if len(b.ToolInput) > 0 {
+			b.ToolInput = pruneZeroValues(b.ToolInput)
+		}
+		b.ThinkingSignature = ""
+		out = append(out, b)
```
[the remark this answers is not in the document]
jasonwc: Good catch — guard added in 3c3f9b5. Now only runs when the field is non-empty, which makes intent explicit and survives the addition of future block types.

pkg/merkle/projection.go
```diff
@@ -0,0 +1,169 @@
...
+			return text[:start]
+		}
+		text = text[:start] + rest[end+len(closeTag):]
+	}
+}
+
+var (
+	trailingLineSpace = regexp.MustCompile(`[ \t]+\n`)
+	consecutiveNewlines = regexp.MustCompile(`\n{2,}`)
+)
```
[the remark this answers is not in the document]
jasonwc: Intentional — this is what catches the harness's "57a58 >" drift case in the PCC-562 evidence. Added a test in 3c3f9b5 that explicitly names this as an accepted tradeoff so a future reader doesn't take the regex out of context and tighten it.

pkg/merkle/node.go
jasonwc: Let's chat about this more at our sync tomorrow.

jasonwc: Since we didn't get a chance to sync, I think we should consider rolling this out as is so that we can see the updated chains, then follow up with another change if necessary.

yeazelm (approved):

pkg/merkle/node.go
jpmcb: > If we just project into bucket, I think that results in loss.

Yes, this essentially was the original idea - project into the `Bucket` which would result in stable hashes - my original intent for `Bucket` was that it would be exactly the projection. Since, in theory, the LLM turns from one JSON payload to the next _should_ be the same, the canonicalized JSON would give us those stable reproducible hashes - but we've clearly established that inference providers and/or harnesses do not hold to that contract so, yes, we need some sort of "normalization" layer.

I'm just not seeing what a layer of indirection from `Bucket` to a normalized `hashableBucket` gets us if `Bucket` should be the projection layer - `Node` is essentially all the metadata and I think it was a bad design on my part to put `Model`, `Provider`, etc. into `Bucket` since they're essentially metadata on the LLM content JSON anyways.

Matt brings up a good point:

>  But I also think we need to make sure we store all the data as it is sent to the model as well.

Then `Node` probably needs a `RawContent json.Raw` field that can keep the exact bits sent from the harness/LLM provider and would _not_fall down into the content-addressable `Bucket`.

---

I'm basically arguing semantics here and "what is a `Bucket`?" - one way or another, this patch gets us there - I do think it's worth considering where that metadata lives and it's probably not in `Bucket` and needs hoisting up into `Node`

PR #214, opened 2026-05-28
✨ feat: add OpenAI embedding support

yeazelm:
## Summary

- Adds an OpenAI-compatible embedding provider behind the existing
  embedding interface.
- Keeps local development on Ollama by default, using `embeddinggemma`
  at 768 dimensions unless OpenAI is selected explicitly.
- Defaults OpenAI embedding config to `https://api.openai.com`,
  `text-embedding-3-small`, and 1536 dimensions when the provider is
  `openai`.
- Lets local embedding services use the key stored by `tapes auth openai`
  when `OPENAI_API_KEY` is not already set.
- Documents the local choice between Ollama and OpenAI embeddings.

## How it works

Embedding config is normalized after flags, environment, and config files
are merged. The selected provider supplies any omitted provider-specific
defaults, so local Ollama and OpenAI do not share a single dimensions or
model default.

The OpenAI provider uses the same `Embedder` contract as Ollama. It
normalizes the configured API base URL, sends embedding requests to the
OpenAI-compatible endpoint, and returns vectors through the existing
embedding path.

For local runs, an explicit `OPENAI_API_KEY` still wins. If it is absent
and OpenAI embeddings are selected, Tapes falls back to the OpenAI
credential stored by `tapes auth openai`.

## Behavior matrix

| Scenario | Before | After |
|---|---|---|
| Default local embeddings | Ollama settings were implicit in shared defaults | Ollama stays explicit: `embeddinggemma`, 768 dimensions |
| OpenAI selected without model/dimensions | Missing provider-specific defaults | OpenAI defaults to `text-embedding-3-small`, 1536 dimensions |
| OpenAI selected with `OPENAI_API_KEY` | No OpenAI embedder path | Environment key is used |
| OpenAI selected without `OPENAI_API_KEY` | No OpenAI embedder path | Stored `tapes auth openai` credential is used when present |

## Test plan

- [x] `go test ./pkg/credentials ./pkg/embeddings/utils ./pkg/embeddings/openai`
- [x] `make format`
- [x] `make test`
- [x] `git diff --check`
- [x] `direnv exec . dagger call build export --path=./build`
- [x] `direnv exec . make test-kafka-e2e`

Refs PCC-579

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,15 +49,33 @@ Install `tapes`:
 curl -fsSL https://download.tapes.dev/install | bash
 ```
 
-Run Ollama and the `tapes` services. By default, `tapes` targets embeddings on Ollama 
-with the `embeddinggema:latest` model - pull this model with `ollama pull embeddinggema`:
+Run the `tapes` services with an embedding provider. Embeddings power `tapes search`;
+choose either local Ollama or OpenAI.
+
+For local embeddings, run Ollama and pull the default `embeddinggemma` model:
 
 ```bash
+ollama pull embeddinggemma
 ollama serve
 ```
+
+Then start Tapes:
+
 ```bash
 tapes serve
 ```
+
+For OpenAI embeddings, store an API key and switch the embedding provider:
+
+```bash
+tapes auth openai
+tapes config set embedding.provider openai
+tapes serve
+```
+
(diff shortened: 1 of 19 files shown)


jasonwc (approved):

jpmcb (approved):
Nice work!

PR #215, opened 2026-05-31
🧹 chore: (pricing) add Claude Opus 4.8 (PCC-618)

bdougie:
## Summary
Assuming this is still needed. I haven't been following the tapes api updates closely as of recent, but in the past I've been manually update the pricing and adding new models as they drop. 

- Add `claude-opus-4.8` to `DefaultPricing()` at $5 input / $25 output / $0.50 cache read / $6.25 cache write (5min) per MTok — same rates as 4.7.
- Add `-4-8` → `-4.8` rewrite in `NormalizeModel` so the canonical Anthropic API ID `claude-opus-4-8` resolves correctly.
- Bump `Last verified` to 2026-05-29.
- Cover the new normalization with a test in `pkg/sessions/summary_test.go`.

Without this, sessions on Opus 4.8 silently report $0.00 cost.

Closes PCC-618.

## Verification

Pricing and model ID confirmed against Anthropic docs:
- Pricing: https://platform.claude.com/docs/en/about-claude/pricing
- Model ID `claude-opus-4-8`: https://platform.claude.com/docs/en/about-claude/models/migration-guide#migrating-from-claude-opus-47
- 4.8 is GA, listed as the current default Opus; 4.7 has moved to the legacy table.

## Out of scope (follow-ups)

- The `Pricing` struct doesn't model 1-hour cache writes ($10/MTok on 4.8) or fast-mode premium ($10 input / $50 output) — same gap that already exists for 4.7/4.6.
- The `inference_geo: "us"` 1.1x multiplier isn't applied anywhere — also pre-existing.

## Test plan

- [x] `go test ./pkg/sessions/...` passes locally
- [ ] `make unit-test` (Dagger) green in CI
- [ ] Manual: send a turn against `claude-opus-4-8-*` through the proxy and confirm the session shows non-zero cost

### DIFF
--- a/pkg/sessions/pricing.go
+++ b/pkg/sessions/pricing.go
@@ -10,7 +10,7 @@ import (
 
 // DefaultPricing returns hardcoded pricing per million tokens for supported models.
 //
-// Last verified: 2026-04-16
+// Last verified: 2026-05-29
 // Sources:
 //   - Anthropic: https://platform.claude.com/docs/en/about-claude/pricing
 //   - OpenAI:    https://platform.openai.com/docs/pricing
@@ -23,6 +23,7 @@ import (
 func DefaultPricing() PricingTable {
 	return PricingTable{
 		// Anthropic
+		"claude-opus-4.8":   {Input: 5.00, Output: 25.00, CacheRead: 0.50, CacheWrite: 6.25},
 		"claude-opus-4.7":   {Input: 5.00, Output: 25.00, CacheRead: 0.50, CacheWrite: 6.25},
 		"claude-opus-4.6":   {Input: 5.00, Output: 25.00, CacheRead: 0.50, CacheWrite: 6.25},
 		"claude-opus-4.5":   {Input: 5.00, Output: 25.00, CacheRead: 0.50, CacheWrite: 6.25},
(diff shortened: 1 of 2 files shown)


jasonwc:
One follow-up worth folding in here (or a fast-follow): `NormalizeModel` gains the `-4-8` → `-4.8` rewrite, but it doesn't strip the 1M-context-window marker that Claude Code sends for the long-context Opus variant — the model id arrives as `claude-opus-4-8[1m]`. That normalizes to `claude-opus-4.8[1m]`, which won't match the new `claude-opus-4.8` pricing key, so **sessions on the 1M variant still report $0.00 cost** even with this PR.

A one-liner before the version rewrites handles it:

```go
// Strip a trailing context-window marker, e.g. "[1m]" on claude-opus-4-8[1m]
if idx := strings.IndexByte(normalized, '['); idx != -1 {
    normalized = normalized[:idx]
}
```

That maps `claude-opus-4-8[1m]` → `claude-opus-4.8` (base price). Note this charges the 1M variant at base Opus rates; Anthropic's >200k-token long-context surcharge isn't modeled, but that's the same per-request-token limitation already called out in your out-of-scope notes — base price is strictly better than $0.

Context: we just hit this downstream — the gateway model allowlists in the console and the clearing were also missing `claude-opus-4-8[1m]` (now fixed), and this is the matching pricing-side gap.

jasonwc (approved):
Approving so I can follow up with Fable pricing.

PR #217, opened 2026-06-04
✨ feat: restructure sessions API — /v1/sessions + /v1/stems (Merkle chains)

jasonwc:
Splits the sessions HTTP API into two surfaces with distinct purposes.

> **Versioning:** this is breaking for API consumers (see the migration table), but tapes is pre-1.0 — bumping as a `0.x` **minor** (`0.11 → 0.12`) per pre-1.0 semver, not a major. The migration table below is the upgrade guide.

## Two surfaces
- **Sessions** — each row represents a harness/agent session (backed by the sessions table populated by the session-tracking ingest path):
  - `GET /v1/sessions` — one session per row, newest-first, cursor-paginated
  - `GET /v1/sessions/:id` — a session plus the selected "stem" of turns and a summary of all stems in the session; `?stem=longest|all`, `?root=<hash>`
- **Stems** — the Merkle leaf-chain / forensic view:
  - `GET /v1/stems` — per-leaf-chain aggregates (cost, tokens, status, duration)
  - `GET /v1/stems/:hash` — a leaf's ancestry chain, root-first
  - `GET /v1/stems/:hash/graph` — graph-shaped DAG projection

A "stem" is a root-to-leaf path of Merkle nodes — what the API previously called a leaf "session". The `tapes deck` TUI, `tapes checkout`, and the API `--web-ui` are updated to the new paths; response shapes for the moved endpoints are unchanged.

## Breaking changes (migration)
| Before | After |
|---|---|
| `GET /v1/sessions` (leaf-node list) | `GET /v1/sessions` now returns one row per harness/agent session; the leaf list is removed |
| `GET /v1/sessions/summary` | `GET /v1/stems` |
| `GET /v1/sessions/:hash` | `GET /v1/stems/:hash` |
| `GET /v1/sessions/:hash/graph` | `GET /v1/stems/:hash/graph` |
| — | `GET /v1/sessions/:id` (new) |

## Docs
- README quickstart now brings up storage with `tapes local up` (Postgres is required — the old flow hit `empty postgres dsn`).
- Fixed the dead "Documentation" link.
- Added swagger annotations for `/v1/sessions` and `/v1/sessions/:id`.

## Test plan
- [x] `make test` (Dagger, Postgres-backed) green
- [x] `--web-ui` endpoints (`/v1/stems`, `/v1/stems/:hash/graph`) verified
- [x] deck + checkout retargeted

refs PCC-635


api/sessions_handlers.go
```diff
@@ -0,0 +1,414 @@
...
+		if n, ok := byHash[hash]; ok {
+			result = append(result, n)
+			for _, kid := range children[hash] {
+				visit(kid)
+			}
+		}
+	}
+	visit(rootHash)
+	return result
+}
```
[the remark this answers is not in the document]
jasonwc: Fixed in de91ceb — `handleGetSession` now validates the id is a UUID up front and returns **400** (the documented response) before calling storage, so a malformed id no longer falls through to 500.

api/sessions_handlers.go
```diff
@@ -0,0 +1,414 @@
...
+		}
+		return 1 + best
+	}
+
+	byHash := make(map[string]*merkle.Node, len(nodes))
+	for _, n := range nodes {
+		byHash[n.Hash] = n
+	}
+
+	var stems []StemSummary
```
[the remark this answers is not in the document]
jasonwc: Fixed in de91ceb — removed the dead `byHash` allocation in `sessionStems`.

pkg/storage/session_record.go
```diff
@@ -0,0 +1,26 @@
...
+
+// SessionRecord is the flat sessions-table row surfaced by
+// GET /v1/sessions. Fields absent in the DB (NULL) are represented as empty
+// strings or nil/zero values so API callers never have to unwrap optional
+// pgtype wrappers.
+type SessionRecord struct {
+	ID                string
+	HarnessID         string
+	HarnessSessionID  string
+	Name              string // empty when not set
```
yeazelm, line 13: FWIW I think we probably want this to be UserSuppliedName and then have an InferredName where the UI can choose to override the InferredName with UserSuppliedName if desired. I think this is fine for now, but we probably need to spend a bit more time honing in on this data model (but that is probably something we can do after this large change lands)

api/api.go
```diff
@@ -69,15 +69,18 @@ func NewServer(config Config, driver storage.Driver, log *slog.Logger) (*Server,
...
+	//   /v1/sessions  — product sessions (sessions-table, UUID-keyed)
+	//   /v1/stems     — Merkle leaf chains (hash-keyed, forensic view)
+	// Static paths are registered before parameterised ones.
 	app.Get("/v1/stats", s.handleStats)
 	app.Get("/v1/sessions", s.handleListSessions)
-	app.Get("/v1/sessions/summary", s.handleListSessionsSummary)
-	app.Get("/v1/sessions/:hash/graph", s.handleGetSessionGraph)
-	app.Get("/v1/sessions/:hash", s.handleGetSession)
+	app.Get("/v1/sessions/:id", s.handleGetSession)
+	app.Get("/v1/stems", s.handleListStems)
```
yeazelm, line 79: I'm still noodling on `stems` for this but I think its fine to get moved in for now and see how we like it.

yeazelm (approved):

api/api.go
jasonwc: Yeah sounds good to me. I like it playing in to our existing naming convention so figured its a good start for now in pre-release.

pkg/storage/session_record.go
jasonwc: Agreed. Good call out.

PR #218, opened 2026-06-09
🔧 fix: capture server_tool_use and web_search_tool_result content

jasonwc:
## Summary
- `server_tool_use` and `web_search_tool_result` content blocks were persisted with only their `type` — the server-tool call input and the web-search results were dropped.
- Capture them end to end: `server_tool_use` reuses the `tool_use` path (id / name / input, streamed via `input_json_delta`); `web_search_tool_result` keeps its `tool_use_id` and the result `content`, stored verbatim.

## Test plan
- [x] `make test` green (adds a reducer spec covering both block types)

Part of PCC-563


jpmcb (approved):

jasonwc:
Addressed the P2 from the review (ParseRequest `server_tool_use` handled only via the generic extraction): `ParseRequest` now has an explicit `if cb.Type == "tool_use" || cb.Type == "server_tool_use"` guard around the id/name/input extraction, symmetric with `ParseResponse`, so a future guard can't silently drop server-tool fields. (dcb18fd)

On the canonical-equivalence fixture gap also noted: tracking as a follow-up — happy to add SSE/one-shot fixture pairs for the two new block types if we want the dedup guarantee locked down here.

jasonwc:
Also added the canonical-equivalence fixture (3972cd3): `turn_04_server_tool_{oneshot.json,stream.sse}` — a server_tool_use + web_search_tool_result turn whose streamed and one-shot forms now assert-equal merkle hashes, locking down the dedup guarantee for the new block types (the `Content json.RawMessage` byte-divergence risk). `go test ./pkg/capture/...` green.

PR #220, opened 2026-06-09
🧹 chore: (pricing) add Claude Fable 5 and strip the [1m] context marker

jasonwc:
Stacked on #215 (rebase onto main once that merges).

Fable 5 is GA but missing from `DefaultPricing()`, so sessions on `claude-fable-5*` report $0.00 cost.

- Add `claude-fable-5` at $10 input / $50 output / $1.00 cache read / $12.50 5min cache write per MTok (Anthropic published pricing).
- Strip the `[1m]` 1M-context marker in `NormalizeModel` so `claude-fable-5[1m]` / `claude-opus-4-8[1m]` resolve to their base pricing keys (1M context is standard pricing, no premium). Without this, 1M-context traffic also prices at $0.00 today.
- Pin the `claude-fable-5` lookup explicitly in `summary_test.go` since the dotted-key round-trip loop skips keys without a minor version.

Related to PCC-663.

### DIFF
--- a/pkg/sessions/pricing.go
+++ b/pkg/sessions/pricing.go
@@ -10,7 +10,7 @@ import (
 
 // DefaultPricing returns hardcoded pricing per million tokens for supported models.
 //
-// Last verified: 2026-05-29
+// Last verified: 2026-06-09
 // Sources:
 //   - Anthropic: https://platform.claude.com/docs/en/about-claude/pricing
 //   - OpenAI:    https://platform.openai.com/docs/pricing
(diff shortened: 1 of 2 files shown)


jasonwc:
Re the P1 about `TrimSuffix("[1m]")` ordering: the suggested reorder is based on a model-ID shape Anthropic doesn't produce. The 1M-context marker always sits at the **very end** of the ID, after any date suffix — `claude-sonnet-4-5-20250929[1m]`, not `claude-fable-5[1m]-20260101`. With the marker trailing, the strip must run **before** the `-YYYYMMDD` stripper (otherwise `[1m]` hides the date from it); moving it after, as suggested, would leave dated+marker IDs un-normalized — the exact regression the comment warns about. Current ordering handles all four real combinations; added a test pinning the dated+marker form in 37f297b.

yeazelm (approved):

PR #227, opened 2026-06-12
🔧 fix: compute per-session cost at read time (related to PCC-686)

isabensusan:
## What

Per-session `total_cost_usd` was always `$0.00` in the console (the session-detail sidebar **and** the sessions table) even though tokens were recorded correctly. That field is a stored column the ingest worker stubs at `0` — no pricing lookup is wired into the proxy (`proxy/worker/pool.go` → `CostUSD: 0, // pricing lookup not wired in this repo`).

This computes per-session cost **at read time** from per-model token rollups via the pricing table — the same approach `/v1/stats` already uses (PCC-452) — instead of trusting the stubbed column.

## Changes

- **`GET /v1/sessions/:id`** (detail / sidebar) — `sessionCostFromNodes` folds the session's nodes into a per-model token map and prices it. Folds over every node in the session, so the cost is stable regardless of `?stem` / `?root`.
- **`GET /v1/sessions`** (list / table) — `SessionTokensByModel` returns a per-`(session, model)` token rollup in one grouped query per page (`ANY($1::uuid[])`, mirroring `getSessionPreviews`); the handler folds each row's cost. A rollup failure is non-fatal — the list falls back to the stored value rather than 500-ing.
- Shared **`s.pricingTable()`** helper; `/v1/stats` and `/v1/stems` now use it too, so all three price identically.
- New Ginkgo specs for `sessionCostFromNodes`.

No API shape change (`total_cost_usd` was always a field); works retroactively for existing sessions.

## Testing

- `make format`, `go build ./...` (`GOEXPERIMENT=jsonv2`), and 60/60 `api` Ginkgo specs pass.
- Verified end-to-end against a local clearing: list + detail report real cost (e.g. `$18.74`) where they previously showed `$0`, and detail/list agree on the same session.
- `SessionTokensByModel` (Postgres) has no local unit test (needs a DB via Dagger); verified against the clearing. Happy to add a postgres integration spec if preferred.

## Notes

Production serves `/v1/sessions` via the tapes-api sidecar (this same Go binary), proxied by paper-cli — so this change ships the fix to production; no paper-cli change is needed.

related to PCC-686

### DIFF
--- a/api/sessions_cost_test.go
+++ b/api/sessions_cost_test.go
@@ -1,0 +1,60 @@
+package api
+
+import (
+	. "github.com/onsi/ginkgo/v2"
+	. "github.com/onsi/gomega"
+
+	"github.com/papercomputeco/tapes/pkg/llm"
+	"github.com/papercomputeco/tapes/pkg/merkle"
+	"github.com/papercomputeco/tapes/pkg/sessions"
+)
+
+// assistantCostNode builds a standalone assistant node carrying a model and
+// token usage — the only fields sessionCostFromNodes reads.
+func assistantCostNode(model string, in, out, cacheCreate, cacheRead int) *merkle.Node {
+	bucket := v1TestBucket("assistant", "ok", model, "anthropic", "")
+	return merkle.NewNode(bucket, nil, merkle.NodeOptions{
+		Usage: &llm.Usage{
+			PromptTokens:             in,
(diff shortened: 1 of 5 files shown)


jasonwc:
Closing in favor of https://github.com/papercomputeco/tapes/pull/228

PR #228, opened 2026-06-15
✨ feat: derive sessions, traces, and spans from an immutable capture layer

jasonwc:
Rebuilds tapes telemetry around a single source of truth: an append-only **`raw_turns`** capture layer, with everything user-facing derived from it by a pure, idempotent deriver. Replaces the node/merkle read surfaces with a span/trace projection.

**The model**
- `raw_turns` — verbatim wire request/response + session envelope per call; immutable, append-only. Every derived row is a pure function of it (re-derive prunes 0).
- A continuous derive worker (`tapes serve derive-worker`) projects raw → merkle nodes → transcript-reconciled tree → **spans/traces/links** (RFD 00007), with deterministic IDs minted from wire identity. Settle-or-max-lag dirty queue, health/readiness probes, metrics, graceful drain.
- Per-session derive-time folds: cost, token rollups (main vs shadow split), turn count, dominant model, **per-model cost-weighted usage breakdown** (`model_usage`), and turn-card previews (user prompt + closing-response answer line, both injection-aware).
- **Span embeddings** + semantic search (`/v1/search/spans`), written by the derive worker when `spec.search` is configured.

**Read surface (v1, node view retired)**
- `GET /v1/sessions` (cursor-paginated; since/until window; auth_subject filter; carries folded model/tokens/cost/turn_count/model_usage), `/v1/sessions/{id}`, `/v1/sessions/{id}/traces` (composite), `/v1/traces*` (lazy summaries → detail → span drill-in), `/v1/stats` (span-grain), `/v1/search/spans`, `/v1/sessions/{id}/raw_turns` (operator wire log).
- **Deleted:** `/v1/stems`, `/v1/stems/:hash/graph` (old DAG), the node-tree endpoint, node-layer `/v1/search`. A DAG remains re-derivable from raw if ever wanted. `tapes chat` removed; `tapes checkout` reimagined as a conversation-export primitive; `tapes search` is span-only; the debug web UI rebuilt on the session surface.

**Stats semantics reviewers should know:** `turn_count` now counts user-visible turns (traces), not wire calls; `total_duration_ns` is the sum of trace durations (agent time), not wall-clock; token/cost are span-rollup sums (delta-only, no re-sent-history inflation); `stem_count` is gone on the span path. Per-node filters (project/agent/model/provider) fall back to the legacy node aggregate.

**Compaction/resume correctness:** compaction is classified structurally (main-thread only) + by the real summary structure (not a bare substring) — a subagent quoting "Primary Request and Intent" no longer false-fires. Resume/model-switch re-sends are bounded by a last-fresh-assistant boundary so replayed history doesn't duplicate traces or bleed into trace content.

**Migrations:** additive only (`1781230000`–`1781290000`), applied on boot; columns backfill via re-derive, not migration. Prod never ingested transcripts; staging re-derives cleanly.

**Response compression** enabled on the API (a full session detail measured 2.65MB raw / ~80KB gzipped).

**Testing:** Dagger suite (~52 packages) + e2e (proxy→ingest→derive→search) green; e2e now runs in CI. Corpus gate pins three live-capture sessions (incl. compaction real+false, resume, nested subagents, 3-model switch) as semantic-equivalence oracles. Generated trees (`gensqlc/`, `docs/`) and corpus fixtures marked `linguist-generated`.

Net diff ~+18k/−6k hand-written (213 files incl. generated/corpus).

related to PCC-687, related to PCC-674, related to PCC-675


migrations/1781137867_derive_queue.up.sql
```diff
@@ -0,0 +1,33 @@
...
+-- clear or a duplicate mark only costs a redundant derive.
+
+CREATE TABLE IF NOT EXISTS derive_queue (
+    org_id             UUID NOT NULL,
+    harness_id         TEXT NOT NULL,
+    harness_session_id TEXT NOT NULL,
+    dirtied_at         TIMESTAMPTZ NOT NULL DEFAULT CURRENT_TIMESTAMP,
+
+    PRIMARY KEY (org_id, harness_id, harness_session_id)
+);
```
jpmcb, line 29: This will fill up really quickly. A future enhancement would probably be to bring on [`pgmq`](https://github.com/pgmq/pgmq) for a durable message queue in postgres vs. a lazy table with `dirtied_at` that fills up.

pkg/derive/chain.go
```diff
@@ -0,0 +1,106 @@
...
+			Role:      msg.Role,
+			Content:   msg.Content,
+			Model:     req.Model,
+			Provider:  call.Provider,
+			AgentName: call.AgentName,
+		}
+		node := merkle.NewNode(bucket, parent, merkle.NodeOptions{
+			Project: call.Project,
+			Request: params,
+		})
```
jpmcb, line 66: Why does the turn's chain need `NewNode` (which computes the content-addressing hash from the parent node)? This looks to take the request/response from the inference provider and compute the entire Merkle DAG which seems counter intuitive if we're using session / trace IDs.

jpmcb (commented):
My biggest question: why leave `pkg/merkle/...` around and how it relates at all to `traces` `turns` `spans` etc. We can ship this, my preference would 100% be to rip those out - no sense in keeping that old data model around.

yeazelm (approved):
Loving the direction here! Let's land this and keep going to switch everything over to it!

migrations/1781137867_derive_queue.up.sql
jasonwc: Yep, we will follow up with better queuing architecture.

pkg/derive/chain.go
jasonwc: Yes, we will follow up with removing persisted nodes table, the legacy node-dag api surfaces, etc... But We still rely on parts of pkg/merkle for the algorithm we use to create content fingerprints for deduping. I think that is a fast follow up of basically pruning dead coded paths, read/write surfaces, and extra persistence we don't need.

PR #229, opened 2026-06-15
✨ feat: version tapes span projection tables

yeazelm:
## Summary
- Move the derived span projection to date-versioned physical tables for the initial `2026-06-15` schema: `span_turns_20260615`, `spans_20260615`, and `span_links_20260615`.
- Add `derived_projection_schemas` metadata so future projection schemas can coexist and old data remains readable.
- Point sqlc queries and span embedding joins at the current versioned table family.

related to PCC-690

## Why this is versioned derived data
The raw tapes turns remain the long-lived source of truth. The span/session tables are a derived visualization model: they encode the current answer to "how should the console understand and render this session?" That answer is allowed to evolve as we learn more from real data and UI needs.

This PR starts the projection at compatibility date `2026-06-15` with date-suffixed physical tables. A future projection bump would be warranted when the derived model changes in a way that old rows cannot safely share the same table contract. For example, if we move from "one trace row per user-visible turn" to a richer conversation graph that changes span identity, link cardinality, branch semantics, or the columns needed to distinguish main-thread work from subagent/synthetic work, we would create a new dated table family, register it in `derived_projection_schemas`, backfill it from raw turns, and move unpinned tenants/users and current reads to the new family.

The older table family is a compatibility snapshot, not the default write target forever. New data should flow to the latest projection unless a tenant/user is explicitly pinned to an older compatibility date. Over time, normal retention and archival policy can age old projection rows out of hot storage, and older table families can become sparse: in the limit, they may only exist for tenants/users that were pinned or had historical data at that projection date. We can also stop auto-creating old projection families for new tenants once no new writes target them.

This is similar in spirit to semantic versioning, but date-based compatibility is a better fit for stored projections. A new dated table family is closer to a major version bump: it means the derived row contract changed enough that old and new rows should not silently share one physical schema. Minor or patch-level changes should stay within the existing projection date when they are compatible, such as adding nullable fields, tuning indexes, improving query plans, regenerating sqlc, or fixing derivation bugs without changing the meaning of existing rows. The date is the compatibility boundary; the table suffix is the physical representation of that boundary.

## Verification
- `direnv exec . make generate`
- `direnv exec . go test ./api ./pkg/derive`
- `direnv exec . go test ./... -run '^$'`
- `direnv exec . make e2e-test`
- `direnv exec . make test-kafka-e2e`
- `direnv exec . make build`
- `git diff --check`

Note: `direnv exec . go test ./api ./pkg/derive ./pkg/spanembed` passes api/derive and reaches the expected local guard for spanembed because `TEST_POSTGRES_DSN` is not set; the Dagger E2E run exercises the Postgres migration path.

### DIFF
--- a/migrations/1781310000_version_span_projection_tables.down.sql
+++ b/migrations/1781310000_version_span_projection_tables.down.sql
@@ -1,0 +1,25 @@
+ALTER TABLE IF EXISTS spans_20260615 RENAME CONSTRAINT spans_20260615_kind_chk TO spans_kind_chk;
+
+ALTER INDEX IF EXISTS span_links_20260615_session_idx RENAME TO span_links_session_idx;
+ALTER INDEX IF EXISTS span_links_20260615_to_trace_idx RENAME TO span_links_to_trace_idx;
+ALTER INDEX IF EXISTS span_links_20260615_from_trace_idx RENAME TO span_links_from_trace_idx;
+ALTER INDEX IF EXISTS span_links_20260615_pkey RENAME TO span_links_pkey;
+
+ALTER INDEX IF EXISTS spans_20260615_org_kind_started_idx RENAME TO spans_org_kind_started_idx;
+ALTER INDEX IF EXISTS spans_20260615_raw_turn_idx RENAME TO spans_raw_turn_idx;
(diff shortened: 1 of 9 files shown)


yeazelm:
@greptile-app try your review again, I accidentally pushed too many commits, now its a much smaller review.

jasonwc (approved):

PR #230, opened 2026-06-16
:wrench: fix: project Codex Responses turns into traces

yeazelm:
## Summary
- preserve Responses API request tools so derived spans can render available tool context
- project Codex tool-role/function-call-output turns and label shell execution spans as Bash
- classify Codex compaction prompts and fix JSONB raw-turn header sizing

## Testing
- direnv exec . go test ./pkg/derive ./pkg/llm/provider/openai
- TEST_POSTGRES_DSN=... direnv exec . go test ./pkg/storage/postgres

related to PCC-690
related to PCC-681

### DIFF
--- a/pkg/derive/classify.go
+++ b/pkg/derive/classify.go
@@ -139,6 +139,8 @@ func ClassifyCall(req *llm.ChatRequest, resp *llm.ChatResponse) string {
 	{
 		lt := strings.ToLower(lastText(req))
 		if strings.Contains(lt, "summary of the conversation so far") ||
+			strings.Contains(lt, "context checkpoint compaction") ||
+			strings.Contains(lt, "create a handoff summary for another llm") ||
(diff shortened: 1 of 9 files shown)


jasonwc (approved):
LGTM! It does make me wonder what the right organization will be eventually:

Should classify contain all logic for determining kind, regardless of provider? Or will we eventually want per model/harness classifiers.

I think the ultimate goal is: we classify and derive everything into our known data model that generalizes these things, but unclear how to cleanly separate provider specific logic in code.

yeazelm:
> Should classify contain all logic for determining kind, regardless of provider? Or will we eventually want per model/harness classifiers.

I agree we probably will want to refactor this into something a bit more generic where classifiers can be tuned for the specifics of the backend but for now this does seem to hold for keeping it together.

PR #234, opened 2026-06-24
🔧 fix: bound derive-worker peak memory with a cgroup-derived soft limit

jasonwc:
## Summary

- A large session's derive re-parses the whole re-sent conversation on every turn, so its transient allocation peak runs ~2× its live set. Under the default GOGC=100 the heap is allowed to roughly double before a collection, so deriving a large session can exceed the container memory limit in a brief spike and get the worker OOM-killed — even though its steady-state live set fits the budget comfortably.
- Fix: the derive worker sets a soft heap limit (`debug.SetMemoryLimit`) derived from its cgroup memory budget (90%, leaving headroom for non-heap memory), so the GC paces allocation against the container budget instead of against live-heap-times-GOGC. No-op when `GOMEMLIMIT` is set by the operator, when no cgroup limit is readable (local dev), or on non-Linux.
- Touches **zero deriver logic**, so re-derive stays byte-identical by construction. On a synthetic 2000-node session the peak drops from ~5.5 GB to ~3.1 GB (its live set) with identical output.
- Scoped to the derive worker; the API, proxy, ingest, and embed workers have different memory profiles.

## How it works

`ApplySoftMemoryLimit` reads the container's cgroup memory limit (v2 `memory.max`, falling back to v1), and sets `debug.SetMemoryLimit` to 90% of it. The runtime then runs the GC more often as the heap approaches the ceiling, capping the transient peak near the live set rather than letting it grow to ~2× before collecting. An operator-set `GOMEMLIMIT` takes precedence (the worker leaves the runtime's value untouched), and an unconstrained or non-Linux environment is a clean no-op.

## Test plan

- [x] Unit tests: cgroup v2/v1 limit parsing + the apply/override/no-op logic
- [x] Synthetic growing-session regression: derives correctly under a tight soft ceiling; plus a `TAPES_MEMPROBE`-gated sampled-peak demonstration
- [x] `make format`, `go vet`, `go build`, derive + worker suites green locally
- [ ] CI: dagger build / e2e / smoke

Fixes PCC-767.

### DIFF
--- a/pkg/derive/worker/memlimit.go
+++ b/pkg/derive/worker/memlimit.go
@@ -1,0 +1,112 @@
+package worker
+
+import (
+	"log/slog"
+	"os"
+	"runtime/debug"
+	"strconv"
+	"strings"
+)
+
+// A full session derive allocates transiently far above its live set:
+// every wire turn re-sends the whole prior conversation, so the deriver
+// re-parses an O(N) request per turn and the per-turn garbage piles up
+// faster than the GC reclaims it. Under the default GOGC=100 the heap is
+// allowed to roughly DOUBLE its live size before a collection, so a big
+// session's transient peak runs ~2x its (already large) live set and can
+// exceed the container memory limit in a brief spike, getting the worker
+// OOM-killed even though its steady-state live set fits the budget
+// (PCC-767).
+//
+// The live set itself fits the budget; only the transient overshoot does
+// not. A soft memory limit (GOMEMLIMIT) is the right tool: it makes the
+// GC pace against a ceiling instead of against live-heap-times-GOGC, so
+// the peak collapses back toward the live set, trading some extra GC CPU
+// for a bounded heap. We derive the ceiling from the cgroup memory limit
+// the orchestrator already sets, so it tracks the container budget
+// without a second knob to keep in sync.
+
+// memLimitFraction is the share of the cgroup memory limit used as the
+// soft heap ceiling. The remainder is headroom for non-heap memory
+// (goroutine stacks, the pgx driver, the Go runtime itself) so the soft
+// limit bites — driving more frequent GC — before the hard cgroup limit
+// OOMKills the process.
+const memLimitFraction = 0.9
+
+// cgroup limit file paths (v2 first, then v1). Overridable in tests.
+var (
+	cgroupV2MaxPath = "/sys/fs/cgroup/memory.max"
+	cgroupV1MaxPath = "/sys/fs/cgroup/memory/memory.limit_in_bytes"
+)
+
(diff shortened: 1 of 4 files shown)


pkg/derive/worker/memlimit.go
```diff
@@ -0,0 +1,109 @@
...
+// OOMKills the process.
+const memLimitFraction = 0.9
+
+// cgroup limit file paths (v2 first, then v1). Overridable in tests.
+var (
+	cgroupV2MaxPath = "/sys/fs/cgroup/memory.max"
+	cgroupV1MaxPath = "/sys/fs/cgroup/memory/memory.limit_in_bytes"
+)
+
+// cgroupUnlimited is the sentinel cgroup v1 writes for "no limit": a
```
yeazelm, line 42: Please please please tell me we aren't on cgroupsv1 anywhere. Is this actually needed or just defense against the now very deprecated v1 settings?

yeazelm (approved):

pkg/derive/worker/memlimit.go
jasonwc: Gonna have it check it out, but I guess this is just a very old defense lol.

jasonwc: 🤖 Good instinct — we're on cgroup **v2** everywhere (stereOS/NixOS uses the unified hierarchy, and the k8s nodes + local Kind are all v2), so the v1 branch never actually fires here. I left it in as cheap graceful-degradation defense: on a host without `memory.max` it just returns "no limit" and the worker runs unbounded exactly as it does today — no misbehavior, just no auto-ceiling. Happy to rip it out if you'd rather keep this lean.

On the bigger question (why didn't Go just do the right thing in k8s before?): the Go runtime doesn't read cgroup limits for memory. Unlike the JVM's `-XX:+UseContainerSupport`, `GOMEMLIMIT` is **opt-in** and the runtime never auto-derives it from the container budget. With nothing set, the GC paces purely against `GOGC` (default 100), which lets the heap grow to ~2× live before collecting — with no awareness of the 2Gi cgroup ceiling. So on a big session the transient re-parse churn overshot the limit and the kernel OOM-killed the pod before the GC had any reason to run; it's not that the GC couldn't reclaim, it just wasn't told to run sooner.

This PR supplies the missing piece: read the cgroup limit and set a soft `GOMEMLIMIT` (90%) so the GC paces against the container budget. The alternatives are wiring the `GOMEMLIMIT` env from the downward API (`resources.limits.memory`) or pulling in a lib like `automemlimit`; doing it in-process keeps it correct-by-default without depending on the env being set — and an operator can still override via the `GOMEMLIMIT` env (now also exposed as a TapesGateway knob). 🤖

PR #236, opened 2026-06-25
🔧 fix: bound derive-worker live memory with clone-on-retain

jasonwc:
## Summary

- The deriver keeps one node per unique conversation turn, but the strings each
  node retains are zero-copy sub-slices of the raw request buffer the provider
  parser returned. Go keeps a backing array alive while *any* sub-slice
  references it, so every retained node pinned its turn's whole multi-MB
  re-sent-history buffer — and because a session re-sends its growing history on
  every turn, the live set grew to ~the re-sent history (quadratic in turns)
  even though the unique content is tiny.
- Fix: clone the retained string/byte surface when a node is first kept, so each
  raw buffer becomes collectable once its turn is processed and the live floor
  settles at ~unique content (linear in turns).
- Cloning copies identical bytes, so node hashes and the entire derived
  projection are unchanged — proven byte-identical against three live-capture
  corpora.
- Durable complement to #234: that paces the GC to collapse the transient *peak*
  toward the live set; this lowers the live *set* itself. Together, peak memory ≈
  O(unique content + structure).

## How it works

When `AddTurn` first retains a node (`!dup`), `Node.CloneRetained` reallocates
every retained string/byte field — bucket type/role/model/provider/agent, every
content block's text / thinking / tool fields / raw JSON, and the node's
hash / parentHash / kind / threadID / parentToolUseID. The request params are
shared across a call's chain, so `System` is cloned once per call. Pinning is
**all-or-nothing** — one surviving alias keeps the whole buffer — so the clone
surface must be complete; the invariant lives in a single helper.

The non-obvious pin was **not** a string: the scalar pointers on `RequestParams`
(`MaxTokens` / `Temperature` / `Stream` / `ToolCount`) are values the provider
request decoder can allocate inside the buffer's arena, so copying the pointer
by value keeps the whole buffer alive. They're repointed at fresh storage.
Response-side `Usage` is parsed by a standalone decoder and measured not to pin,
so it's left as-is.

## Memory impact (measured)

Streaming each raw row and dropping it (production worker shape), post-GC live floor:

| session | re-sent wire | floor before | floor after | reduction |
|---|---|---|---|---|
| `cb9a87e5` — 87 turns, plan mode + 2 subagents | 14.8 MB | 20.9 MB | 3.5 MB | 6.0× |
| `9fec0da7` — 121 turns, compaction + multi-model | 26.9 MB | 38.2 MB | 4.8 MB | 7.9× |
| `0440f43d` — 108 turns, 19 sessions | 28.1 MB | 36.2 MB | 4.9 MB | 7.4× |
| synthetic — 400 turns | 203 MB | 350.6 MB | 7.6 MB | 46× |

The reduction grows with session size, because the floor drops from O(N²) (every
retained node pinning its turn's full re-sent history) to O(N) (unique content).
Composed with #234's GC pacing, the synthetic's transient peak drops on top of
the floor collapse (655 MB → 359 MB at 400 turns).

## Test plan

- [x] Byte-identical re-derivation gate over three live-capture corpora: full
      node identity/structure + the entire `EmitSpans` projection (Input/Output
      content blocks, links, cross-trace compaction seams), against goldens
      frozen from pre-change `main`
- [x] Deterministic post-GC live-floor assertion (synthetic): floor ≪ re-sent
      wire sum, guarded in CI; sampled-peak demo stays behind `TAPES_MEMPROBE`
- [x] Unit tests: clone helpers preserve every byte while breaking every alias to
      the source buffer (incl. `ToolInput` map rebuild + scalar-pointer realloc)
- [x] `make format`, `make check` (generate / go-mod-tidy / lint / test), `make test` green locally
- [ ] CI: dagger build / e2e / smoke

Related to PCC-767


yeazelm (approved):
Nice find!

pkg/derive/deriver.go
```diff
@@ -254,9 +254,27 @@ func (dv *Deriver) AddTurn(rec *storage.RawTurnRecord) {
...
+	// Params() per call); clone its System once per call, keyed by pointer
+	// identity, rather than re-cloning the system prompt for every node.
+	var srcParams, clonedParams *llm.RequestParams
 	for _, node := range chain {
 		retained, dup := dv.byHash[node.Hash]
 		if !dup {
+			if rp := node.Request; rp != srcParams {
+				srcParams, clonedParams = rp, rp.Clone()
+			}
+			node.CloneRetained(clonedParams)
```
jpmcb, line 277: I understand that by lifting/shifting bits from the `chain` into a zero-copy slice, we can release the memory in the derive worker from those original requests to the garbage collector.

But what's not clear to me is why this design is better vs. getting `rederiveChain` to piece meal a super long chain vs. load the entire `chain []*merkle.Node` - seems the core problem here is we expect the derive worker to load and project huge amounts of content all at once.

Something we should ask is if deriving the entire text and projection of a chain is necessary vs. having seperate APIs for rendering the content of a chain.

Issue #94, opened 2026-02-13
AI Insights: reliable Ollama-backed facet extraction for deck analytics


bdougie:
## Summary

The AI Insights panel in `deck --web` analytics is currently disabled. The Ollama-backed facet extraction has reliability and performance issues that need to be resolved before re-enabling.

## Problems

- Ollama responses often fail JSON parsing (`invalid character 'O' looking for beginning of value`) even with `format: "json"` set
- Backfilling all sessions blocks the `/api/facets` endpoint, making the analytics page load extremely slow
- Auto-detecting Ollama on startup adds latency even when insights aren't needed

## Current State

- UI is commented out in `deck.js` (`analyticsInsightsEl.hidden = true`)
- Server-side facet endpoints still exist and work with `--insights` flag
- `facets_ollama.go` has `format: "json"` set but models still return non-JSON

## Proposed Work

- Investigate async/background backfill so it doesn't block page load
- Add retry logic or better prompt engineering for JSON extraction
- Consider streaming partial results to the UI as sessions are processed
- Re-enable the insights UI once extraction is reliable

<img width="1123" height="287" alt="Image" src="https://github.com/user-attachments/assets/7fbd2e9f-6309-4621-a125-8826fbcad4a8" />

PR #243, opened 2026-07-07
✨ POC: detect session outcomes at derive time

bdougie:
Sessions now carry the artifacts they produced — **pull requests, repos, GitHub issues, Linear issues** — folded from tool spans at derive time. Implements [PCC-840](https://linear.app/paper-compute-co/issue/PCC-840/outcome-model-matcher-registry-in-tapes), the keystone of the [PCC-837](https://linear.app/paper-compute-co/issue/PCC-837/explore-outcomes-as-a-feature-detecting-whether-a-session-produced) outcomes exploration ("did this long/expensive session produce anything?").

## Design

- **Matcher registry** (`pkg/sessions/outcomes.go`), two families:
  - **Bash/`gh`**: `gh pr create` / `gh repo create` / `gh issue create`, gated on the command string, artifact URL parsed from the tool output. Conservative: no URL in output (errored/interrupted create) → no outcome. `gh pr view` printing a URL doesn't count.
  - **MCP**: Linear issue-create tools (`*linear*save_issue`/`create_issue`), URL + title from the result payload, JSON-first with a bare-URL fallback.
  - `kind` is an open set; adding a detected artifact type is one table row.
- **Fold mirrors `model_usage`**: `EmitSpans` folds per-session outcomes (deduped by URL, capture order) into the `SpanSet`; `writeSpanSet` persists `sessions.outcomes JSONB`; `SessionRecord` → `SessionItem` expose it on `GET /v1/sessions[{id}]`.
- **Provenance**: each outcome carries `trace_id`/`span_id` of the detecting tool span — deliberately matching the simulacrum memory stream's source-pointer convention (pcc-labs/tapes-simulacrum) so a future observation/dream can cite "this session shipped PR X" with evidence. `detected_at` is the detecting span's start time, **never the wall clock**, so re-derive reproduces the fold byte-for-byte.
- **No backfill by design** (PCC-837 decision): pre-migration rows carry NULL until their next organic derive pass; the admin re-derive endpoint can opt a session in.

## Consumers already speaking this shape

- `paper recap <trace_id>` CLI + paperd (control-proto v17) decode `session.outcomes` with tolerant defaults — outcomes light up there the moment this merges.
- Console `OutcomesSection`/`RecapCard` (console.papercompute.com#146/#147) render the same wire shape, including the explicit "No outcome recorded" empty state.

## Test plan

- `go test ./pkg/sessions/ ./pkg/derive/ ./api/` — new Ginkgo specs cover every matcher (success, errored create, case-insensitivity, non-create commands, JSON + bare-URL Linear payloads, non-issue Linear tools), URL dedup, and the derive fold's provenance stamping + cross-turn dedup + empty-fold case.
- Postgres integration suites run in CI via the Dagger services environment (`make test`) — locally confirmed pre-existing-green packages stay green; `sqlc generate` + `make swag` outputs committed.
- Migration follows the `derived_model`/`model_usage` additive-JSONB pattern; down migration drops the column.

part of PCC-837 · implements PCC-840

https://claude.ai/code/session_01Moncs7rC9gx5gWnCNgWyz5

### DIFF
--- a/api/sessions_handlers.go
+++ b/api/sessions_handlers.go
@@ -57,7 +57,14 @@ type SessionItem struct {
 	// count, so the UI can show "dominant model + per-model %" without a
 	// cheap-subagent fan-out skewing it. Populated on the session detail;
 	// nil until the session first derives.
-	ModelUsage      []ModelUsage   `json:"model_usage,omitempty"`
+	ModelUsage []ModelUsage `json:"model_usage,omitempty"`
+	// Outcomes is the fold of artifacts the session produced (pull
+	// requests / repos / issues), detected from tool spans at derive
+	// time and deduped by URL. Each carries trace/span provenance back
+	// to the detecting tool call. Nil until the session derives or when
+	// it produced nothing — "no outcome" is a signal the UI renders
+	// explicitly (PCC-837/PCC-840).
+	Outcomes        []Outcome      `json:"outcomes,omitempty"`
 	HarnessMetadata map[string]any `json:"harness_metadata,omitempty"`
 	Preview         string         `json:"preview,omitempty"`
 	// AuthSubject is the gateway-stamped JWT subject (WorkOS user id)
@@ -75,6 +82,21 @@ type ModelUsage struct {
 	InputTokens  int64   `json:"input_tokens"`
 	OutputTokens int64   `json:"output_tokens"`
 	CostUsd      float64 `json:"cost_usd"`
+}
+
+// Outcome is one artifact a session produced, in the API: kind is an
+// open set (pull_request / repo / issue / linear_issue today), url is
+// the artifact's identity, and trace_id/span_id point back at the
+// detecting tool span so consumers can deep-link the exact turn.
+type Outcome struct {
+	Kind       string    `json:"kind"`
+	URL        string    `json:"url"`
+	Title      string    `json:"title,omitempty"`
+	Repo       string    `json:"repo,omitempty"`
+	TraceID    string    `json:"trace_id,omitempty"`
+	SpanID     string    `json:"span_id,omitempty"`
+	DetectedBy string    `json:"detected_by,omitempty"`
+	DetectedAt time.Time `json:"detected_at,omitzero"`
 }
 
 // SessionListResponse is the response envelope for GET /v1/sessions.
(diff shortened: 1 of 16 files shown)


bdougie:
**Design decision worth flagging: outcome provenance aligns with the simulacrum memory stream.**

Each detected outcome carries `trace_id` / `span_id` pointing back at the detecting tool span, and `detected_at` is that span's start time (never the wall clock). This deliberately mirrors the source-pointer convention in pcc-labs/tapes-simulacrum (`source_session_id` / `source_trace_id` / `source_span_id` on the `simulacrum` table) so the two layers compose:

- an **observation extractor** can read `SessionItem.outcomes` and emit a high-importance memory like *"session X shipped PR #94"* with the outcome as its `evidence`, citing the exact span;
- the **dream digest** gets concrete, deterministic highlights (merged PRs, opened issues) instead of having to infer them from prose — outcomes are a high-signal *input* to the stream, not a competing store;
- both layers stay **rebuildable derived projections** over immutable telemetry: this fold re-derives byte-for-byte (deterministic `detected_at`), same principle as "simulacrum is a derived, disposable memory layer."

Structured artifact detection stays here on the session row (deterministic, matcher-based); natural-language memory stays in the simulacrum (LLM-synthesized, embedded). The shared provenance vocabulary is the bridge.

bdougie:
**CI note: the red `tapes:test` here is not this PR's code.**

This branch builds clean and its own suites pass (`go build ./...`, `go vet`, `pkg/sessions` + `pkg/derive`, and the `pkg/storage/postgres` suite all green locally against the pg_duckdb image). The failure is a **pre-existing flaky test on `main`**: `pkg/storage/postgres` → "orders the list by most-downloaded when asked".

Root cause: the test sets `DownloadCount` on the record and upserts, but `UpsertSkill` never writes `download_count` (that column is owned by the `IncrementSkillDownloads` counter). Both rows persist `download_count = 0`, so `ORDER BY download_count DESC, id DESC` breaks the tie by random UUID → ~25–30% coin-flip failure of `tapes:test` on **any** PR (measured 5/20 on main, 6/20 here, fresh isolated DBs). #243 just drew the short straw twice; #241/#242 got lucky.

Fix is up in **#245** (tracked as **PCC-854**). To turn this PR green: merge #245, then merge/rebase `main` in here. Re-running before that is still a coin flip.

Also noted: #241's description flagged a *second* intermittently-flaky spec in the same suite (`GetSessionRecordByHarness`) — separate follow-up, tracked under PCC-854.

bdougie:
Closing per the 2026-07-15 Dreams rescope: this POC proved outcome detection is feasible and its findings fed the PRD. The product ask is now PCC-904 (journey-level); engineering owns the implementation and may take a different approach. Kept as prior art.

PR #244, opened 2026-07-07
✨ feat: (api) add GET /v1/stats/skills skill-invocation aggregate (PCC-852)

bdougie:
## What

Adds the read-side aggregate for skill-usage insights ([PCC-852](https://linear.app/paper-compute-co/issue/PCC-852/skill-insights)): `GET /v1/stats/skills?since&until` returns skill invocations grouped by skill name for the org's window, ordered by invocation count.

```json
{
  "skills": [
    {"skill": "grove-refresh", "invocations": 12, "error_count": 1,
     "session_count": 8, "completed_session_count": 6,
     "last_used_at": "2026-07-07T12:00:00Z"}
  ],
  "total_invocations": 15
}
```

## How

Skill invocations are already captured: a Claude Code skill call is a tool span with `name='Skill'` and the skill name at `input->0->'tool_input'->>'skill'` (`pkg/derive/spans.go` emit path). This PR adds no capture or derive changes — detection is query-time only, so it works over all history retroactively.

- **sqlc query** `AggregateSkillUsage` (`queries/spans.sql`): groups Skill tool spans by skill name; per-skill invocations, error count (span `status='error'`), distinct sessions, sessions with `derived_status='completed'`, and last-used timestamp. Counts are per unique `tool_use` — the span layer dedups re-sent history, so raw-turn replay never inflates them.
- **storage**: `SkillUsage` + `SkillUsageReader` capability interface (`span_model.go`), implemented by the postgres driver. Span-projection backends only; there is no node-layer fallback, so non-capable backends get 501 from the handler.
- **api**: `handleSkillUsageStats` + route, following the `/v1/stats` handler shape (org from context, RFC3339 `since`/`until`). Swagger regenerated.

Skill names are the strings harnesses invoked and are **not** guaranteed to match a `/v1/skills` record — consumers should join on slug and tolerate unmatched names (that unmatched set is itself the "off-paper skills" discovery signal). The durable skill-ID association is follow-up work on PCC-852.

## Tests

- Handler specs: 501 on non-capable backend, response shape + `total_invocations` fold, empty-list (not null), since/until pass-through, 400 on malformed timestamps.
- Postgres integration specs drive the **real pipeline** — `PutRawTurn` (wire-shaped raw request/response with Skill + Bash tool_use blocks) → `RederiveFromRaw` → `AggregateSkillUsage` — asserting grouping, ordering, org isolation, window filtering, non-Skill tool exclusion, and idempotency across re-derives. This also pins the emitter contract the feature depends on (Skill tool_use → span named `Skill` with `tool_input.skill`).
- Ran against local pgvector postgres: 93/94 suite specs pass. The one failure is the pre-existing flaky `skills_test.go` "orders the list by most-downloaded" spec — it fails intermittently in full-suite runs on unrelated commits and passes 3/3 focused; not touched by this change.

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -80,6 +80,7 @@ func NewServer(config Config, driver storage.Driver, log *slog.Logger) (*Server,
 	// span projection beneath them. Static paths are registered before
 	// parameterised ones.
 	app.Get("/v1/stats", s.handleStats)
+	app.Get("/v1/stats/skills", s.handleSkillUsageStats)
 	app.Get("/v1/sessions", s.handleListSessions)
 	app.Get("/v1/sessions/:id/traces", s.handleGetSessionTraces)
 	app.Get("/v1/sessions/:id/raw_turns", s.handleListSessionRawTurns)
--- a/api/v1_handlers.go
+++ b/api/v1_handlers.go
@@ -124,6 +124,70 @@ func (s *Server) handleStats(c *fiber.Ctx) error {
 	})
 }
 
+// SkillUsageItem is one skill's invocation rollup in the
+// /v1/stats/skills response. Skill is the name string the harness
+// invoked (tool_input.skill) — it is not guaranteed to match a skills
+// record; consumers join on slug and tolerate unmatched names.
+type SkillUsageItem struct {
+	Skill                 string    `json:"skill"`
+	Invocations           int       `json:"invocations"`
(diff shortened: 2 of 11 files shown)


bdougie:
This was an experiment. I think we will need to think a bit on this. Issue is still open.

PR #246, opened 2026-07-09
✨ feat: export session data over HTTP (single session + 30-day bundle)

jasonwc:
## Summary

- Two org-scoped NDJSON export endpoints, backing the console's new "export my data" feature.
- `GET /v1/sessions/:id/export` — one session's turns as JSONL.
- `GET /v1/sessions/export?since=&until=` — the org's sessions in a window as one streamed JSONL bundle (default trailing 30 days).
- Extracts the existing `tapes checkout` JSONL serializer into a shared `pkg/export`, so the CLI and the API emit the identical per-turn record shape (CLI output byte-identical, verified by golden test).

## How it works

- `pkg/export.SessionJSONL(ctx, querier, sessionID, w io.Writer)` writes one JSON object per turn (the existing `exportTurn` shape) to any writer — generalized from the CLI's `*strings.Builder` to `io.Writer` so the API can stream.
- `handleExportSession` mirrors `handleGetSession`/`handleDeleteSkill`: capability-check → 501, org-scoped `GetSessionRecord` → 404 (no cross-org leak), NDJSON headers + `Content-Disposition`, stream the session.
- `handleExportSessions` streams via fasthttp `SetBodyStreamWriter`, iterating `ListSessionRecords` pages past the 200-row UI cap with the existing keyset cursor, writing JSONL as it goes — memory is bounded by page size, not bundle size. `since` is clamped to `now−30d` so the 30-day window is the maximum span (R-20). Registered before `/:id` to avoid route shadowing.
- No schema/migration change; no new auth boundary (org scope via the existing trusted `X-Tapes-Org-Id`).

## Test plan

- [x] `pkg/export` golden test — extracted serializer byte-identical to pre-refactor checkout
- [x] `GET /:id/export` — 200 + full JSONL, filename, cross-org 404, malformed-id 400, 501 gate
- [x] `GET /export` — default 30-day window, since/until override, org isolation, >200-row cap bypass, chunked streaming, route-shadowing
- [x] `since`-clamp — a 40-day-old session is excluded even with `since=1970`
- [x] gzip-over-streaming — `Accept-Encoding: gzip` yields a correctly compressed streamed NDJSON body
- [x] `go build ./...`, `go vet`, `gofmt` clean; validated end-to-end in a clearing against real Postgres

Design + requirements + full test plan live in `docs/features/0001-export-data-from-console/` (companion console PR) and the Linear project.

Non-blocking follow-ups: redundant per-session org re-check (N+1) in the bundle loop; stream duration under-reported by the RED-metrics middleware for streaming responses.

Part of PCC-858

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -81,6 +81,10 @@ func NewServer(config Config, driver storage.Driver, log *slog.Logger) (*Server,
 	// parameterised ones.
 	app.Get("/v1/stats", s.handleStats)
 	app.Get("/v1/sessions", s.handleListSessions)
+	// /v1/sessions/export has no :id segment, so it must be registered
+	// before the bare /v1/sessions/:id route below — otherwise Fiber would
+	// match it as :id="export" and never reach this handler.
+	app.Get("/v1/sessions/export", s.handleExportSessions)
 	app.Get("/v1/sessions/:id/traces", s.handleGetSessionTraces)
 	app.Get("/v1/sessions/:id/raw_turns", s.handleListSessionRawTurns)
 	app.Get("/v1/sessions/:id/export", s.handleExportSession)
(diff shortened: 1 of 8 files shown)


jasonwc:
Superseded — reopening under the Gardener GitHub App identity. This PR was opened as a human operator by mistake; the branch commits have been re-authored to gardener[bot] and the PR is being recreated by the app so the human reviews/approves work they did not author. Same branch, same code.

PR #249, opened 2026-07-11
✨ feat: nest session exports with trace/span granularity

jasonwc:
## Summary

- Reworks both export endpoints (`GET /v1/sessions/:id/export`, `GET /v1/sessions/export?since=&until=`) from one flat JSONL line per turn to **one line per session**, nesting each session's traces and each trace's full spans.
- Each spans-grain line is byte-identical to `GET /v1/sessions/{id}/traces?payload=full`, rendered through a shared `exportSessionLine` helper — a bulk export is exactly a concatenation of single-session exports, and the export surface cannot drift from the session-detail API.
- Adds `?detail=traces|spans` to both endpoints: `spans` (default) carries the full projection; `traces` exports turn headers only, read via `ListTraceSummaries` so span payloads are never loaded server-side. Header-only lines omit `spans`/`links`/`tasks`/`kind_counts` keys entirely (absent = "not exported at this grain", never "empty"). Unrecognized values 400.
- `tapes checkout --format jsonl` intentionally keeps its flat transcript grain; `pkg/export` is now checkout-only.

## How it works

The handlers now read the span model directly (`ListSessionSpanModel` / `ListTraceSummaries`) instead of going through the skill querier's narrowed span view, which carries no payloads or usage. The single-session path still buffers before committing attachment headers (clean JSON errors on failure); the bulk path still streams via `SetBodyStreamWriter` with keyset paging past the 200-row list cap, flushing per session. Org-scoped 404 preflight, dual 501 capability gates, and the 30-day window clamp are unchanged.

## Behavior

| Request | Before | After |
|---|---|---|
| `/v1/sessions/{id}/export` | N flat turn lines | 1 nested session line (full spans) |
| `/v1/sessions/{id}/export?detail=traces` | n/a | 1 session line, turn headers only, `-traces.jsonl` filename |
| `/v1/sessions/export` | turn lines across sessions | 1 nested line per session, streamed |
| `/v1/sessions/export?detail=bogus` | n/a | 400 |

## Test plan

- [x] `make test` (Dagger) — export suites rework: nested-line shape asserted by unmarshalling into `SessionTracesResponse`; traces-grain key omission; 400s; org isolation; paging/clamp/gzip/streaming unchanged
- [x] `make e2e-test` (Dagger)
- [x] Live validation in a grove clearing: real captured Claude session exported at both grains, old-vs-new comparison against the v0.22.0 release image
- [ ] Console download surface (separate PCC-858 follow-up)

Related to PCC-858

### DIFF
--- a/api/sessions_export_all_handlers_test.go
+++ b/api/sessions_export_all_handlers_test.go
@@ -231,6 +231,29 @@ var _ = Describe("GET /v1/sessions/export", func() {
 		Entry("when until is before since", now.Add(-2*time.Minute), now.Add(-5*time.Minute)),
 	)
 
+	// detail=traces: one header-only line per session, grain in the filename.
+	It("exports turn headers per session at detail=traces", func() {
+		drv := newPagingDriver(org, 3, now)
+		server := newExportServer(drv)
+
+		resp, body := getRaw(server, "/v1/sessions/export?detail=traces", org)
+		Expect(resp.StatusCode).To(Equal(fiber.StatusOK))
+		Expect(resp.Header.Get("Content-Disposition")).To(ContainSubstring("-traces.jsonl"))
+
+		lines := nonEmptyLinesAPI(string(body))
+		Expect(lines).To(HaveLen(3))
+		Expect(string(body)).To(ContainSubstring(`"trace_id":"session-0000-t1"`))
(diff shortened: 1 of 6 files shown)


yeazelm (approved):
Besides the greptile feedback, this looks good!

PR #251, opened 2026-07-13
🔧 fix: sanitize jsonb-invalid Unicode on ingest and classify 22P05 as 4xx

jasonwc:
## What

Payloads that contain an escaped NUL (`\u0000`), a raw NUL byte, or an unpaired UTF-16 surrogate are valid JSON but cannot be stored in a Postgres `JSONB` column — the insert fails with `SQLSTATE 22P05`. Since `raw_turns.raw_request`, `nodes.bucket`, and `nodes.content` are all `JSONB`, such a payload failed on every write path, and on the synchronous transcript ingest endpoint it surfaced as a 502.

This PR:

1. **Sanitizes** those sequences to `U+FFFD` before the insert (`sanitizeJSONB`), delegating escape/surrogate/UTF-8 handling to `encoding/json/jsontext` rather than a hand-rolled parser. The trigger is narrow: clean payloads and benign escapes are returned unchanged and unallocated; only an escaped NUL or a surrogate escape takes the decode/re-encode path. Applied to `raw_turns` and to node `bucket`/`content`.
2. **Classifies** `22P05`/`22021` as `storage.ErrInvalidContent`, so a malformed payload returns **422** (retrying the identical bytes can't help) instead of a 502 that reads as a gateway fault.
3. **Meters** the transcript ingest path on `tapes_ingest_writes_total` (`provider="transcript"`), which was previously unmetered — so its writes were invisible on the success counter.

Node hashes are computed on the canonical struct and dedup keys on the pre-sanitized bytes, so rewriting stored JSON changes neither.

## Why

A malformed payload was failing `JSONB` inserts and returning 502s rather than a 4xx, and the transcript write path wasn't reflected in the write-success metric — so the failures didn't show up on the counter.

## Verification

- `go build` / `go vet` / `gofmt` clean.
- Unit specs for each rewritten sequence, value preservation (numbers/booleans/null/arrays, valid surrogate pairs), and the escaped-backslash false positive.
- Handler specs: unstorable content → 422, genuine storage fault → 502, accepted path metered.
- DB-integration spec against Postgres 17: an escaped-NUL payload now stores (scrubbed) instead of raising 22P05.
- Sanitization is allocation-free on clean payloads; only an offending payload pays the decode/re-encode.

## Follow-up (not in this PR)

- Derive-worker quarantine, so a persistently-unstorable row isn't re-attempted on every poll.

related to PCC-870

---

### DIFF
--- a/ingest/ingest.go
+++ b/ingest/ingest.go
@@ -387,6 +387,14 @@ type transcriptMeta struct {
 	Records     int    `json:"records"`
 }
 
+// transcriptWriteProvider labels transcript-sourced writes on the shared
+// tapes_ingest_writes_total counter. The counter's "provider" dimension is
+// a wire-capture notion; a transcript has no LLM provider, so it carries
+// this sentinel instead of falling into the "unknown" bucket it would
+// otherwise share with malformed wire turns. Dashboards select
+// provider="transcript" to see this path's health in isolation.
+const transcriptWriteProvider = "transcript"
+
 // handleTranscriptIngest appends one transcript file to the raw layer.
 // Idempotent per content version: the dedup key includes a content
 // hash, so re-uploading an unchanged file is a no-op while a grown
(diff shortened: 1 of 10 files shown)


jasonwc:
Addressed the one gap flagged in the Greptile review: the two `json.Marshal` exits in `handleTranscriptIngest` (meta + session envelope) now record `tapes_ingest_writes_total{provider="transcript",status="internal_error"}` before returning 500, so every handler exit is metered — closing the visibility gap noted. Pushed in `1a1934b`.

Also folded in fixes for the CI lint findings in the new test files (dupword / gci / intrange / perfsprint) so `check-lint` is green locally against golangci-lint v2.8.0.

yeazelm (approved):

PR #252, opened 2026-07-13
✨ feat: publish OpenAPI 3.0.3 contract

jasonwc:
## What

Make tapes publish a real, versioned **OpenAPI 3.0.3** document (`api/openapi.yaml`) as its contract, so downstream services can generate typed clients from it. The paper CLI's `paper sessions {list,get,export}` (a companion PR) is the first consumer.

related to PCC-872.

## Why

tapes has only ever published a Swagger **2.0** doc (swaggo/swag). Typed-client generators used downstream — notably the paper CLI's `progenitor` + `openapiv3`, which are **OpenAPI 3.0.x only** (not 3.1) — cannot consume 2.0.

## How

`make openapi` runs swag to produce `docs/swagger.json`, then `cmd/gen-openapi` converts it to 3.0.3 in-process with `kin-openapi/openapi2conv` (no Node toolchain) and writes `api/openapi.yaml`.

Two constraints drove the converter:

- **operationIds are mandatory.** progenitor hard-panics on any operation missing an `operationId`, and swag doesn't emit them unless annotated. Every handler now carries an `@ID` with a clean consumer-facing name (`listSessions`, `getSession`, `exportSession`, …), and the converter also synthesizes a deterministic id for any operation still missing one, so a future un-annotated handler can't break downstream codegen.
- **schema names.** swag qualifies schema names with their Go package (`api.SessionItem`, `github_com_..._pkg_llm.ErrorResponse`), which turns into noisy generated type names. The converter strips the package prefix and rewrites every `$ref`, keeping the qualified name only when two packages would collapse onto the same bare name **or** when the bare name would shadow a Rust prelude type (e.g. `Result`) and break a generated client.

The Scalar viewer now serves the embedded 3.0.3 spec at `/swagger/openapi.yaml` instead of swag's 2.0 JSON, so the rendered reference and any vendored client are the same document.

## Verify

- `make openapi` → `api/openapi.yaml` with `openapi: 3.0.3` and an `operationId` on every one of the 14 paths / 17 operations; reproducible (byte-identical on re-run).
- `make test` (Ginkgo) green, including new unit tests for the converter's prefix-strip + collision/reserved-name guards.


yeazelm (approved):

go.mod
```diff
@@ -30,6 +31,7 @@ require (
 	github.com/swaggo/swag v1.16.4
 	golang.org/x/mod v0.32.0
 	golang.org/x/term v0.40.0
+	sigs.k8s.io/yaml v1.6.0
```
jpmcb, line 34: Why k8's yaml package? From the last time I used this lib, it was rather non-intuitive and very k8s contributor focused without newer features. I saw the comment on the output from `sigs.k8s.io/yaml` marshaling to json to then unmarshal to yaml resulting in sorted keys but I'm not seeing why that's necessary if we're going to be generating the openapi yamls every time anyways.

cmd/gen-openapi/main.go
```diff
@@ -0,0 +1,305 @@
+// Command gen-openapi converts the swaggo-generated Swagger 2.0 document
+// (docs/swagger.json) into a clean OpenAPI 3.0.3 document written to
+// api/openapi.yaml.
```
jpmcb, line 3: I'm not a fan of this being a hand rolled transformer from swag --> openapi since there's _alot_ of complexity and 1000 little edge cases here. Looks like there's a long standing issue in swaggo for supporting openapi v3 https://github.com/swaggo/swag/issues/386 so maybe this is complexity we need to consume for now.

cmd/gen-openapi/main.go
```diff
@@ -0,0 +1,305 @@
...
+	"Result": true, "Option": true, "String": true, "Vec": true, "Box": true,
+	"Self": true, "Ok": true, "Err": true, "Some": true, "None": true,
+	"Iterator": true, "Future": true,
+}
+
+// stripSchemaPrefixes renames component schema keys from swag's package-qualified
+// form (api.SessionItem, github_com_..._pkg_llm.ErrorResponse) to their bare type
+// name (SessionItem, ErrorResponse) and rewrites every $ref accordingly. A
+// qualified name is left untouched when its bare form would either collide with
+// another schema's bare form or shadow a reserved Rust prelude name.
```
jpmcb, line 228: This essentially implements a full openapi reference parser (expanding `$ref` tags down into their `path.to.component`). This is aton of complexity to consume and we should aim to get off a hand rolled OpenAPI transformer ASAP.

makefile
```diff
@@ -39,6 +39,10 @@ swag: ## Runs the swaggo/swag utility for generating the swagger yaml
 swagfmt: ## Runs swaggo/swag for formatting swag godoc comments
 	swag fmt
 
+.PHONY: openapi
+openapi: swag ## Regenerates the OpenAPI 3.0.3 contract (api/openapi.yaml) from the swag docs
+	go run ./cmd/gen-openapi
```
jpmcb, line 44: Ideally this would also land something in Dagger's CI so we error when generation has drifted - see the `CheckGenerate` check: https://github.com/papercomputeco/tapes/blob/f60ef393195163cf4ba26da56efe80948eca509c/.dagger/main.go#L55

api/swagger_middleware.go
```diff
@@ -1,13 +1,18 @@
...
+	_ "embed"
 
-	// Register the generated OpenAPI spec with swag.ReadDoc.
-	_ "github.com/papercomputeco/tapes/docs"
+	"github.com/gofiber/fiber/v2"
 )
 
+// openAPISpec is the versioned OpenAPI 3.0.3 contract, generated from the swag
+// docs by `make openapi` (cmd/gen-openapi). It is the same document paper vendors
+// to generate its Rust client, so the viewer and the client always agree.
```
jpmcb, line 11: We may have many clients! No need to reference Rust

api/swagger_middleware.go
```diff
@@ -18,21 +23,25 @@ const scalarHTML = `<!doctype html>
     <meta name="viewport" content="width=device-width, initial-scale=1" />
   </head>
   <body>
-    <script id="api-reference" data-url="/swagger/doc.json"></script>
+    <script id="api-reference" data-url="/swagger/openapi.yaml"></script>
```
jpmcb, line 26: This effectively removes the need for a base `doc.json`: we don't have any other consumers of it besides this POC Scalar page.

So, if we're totally aligning on OpenAPI as the source of truth and specification and don't need a `gen swag --> swag doc --> transform to openAPI --> openAPI doc` pipeline, then we should just use swaggo's v2 release candidate: https://github.com/swaggo/swag/releases/tag/v2.0.0-rc5

```
swag init --v3.1
```

I would much rather stick to `swag` generation then have to consume the complexity of parsing their generated doc

cmd/gen-openapi/main.go
jasonwc: Yeah thats correct. Swaggo does not currently support 3.0.3 and on the rust side we use progenitor which only support 3.

makefile
jasonwc: Thats a good followup. Adding a ticket.

PR #257, opened 2026-07-19
🔧 fix(api): snap and cache /v1/stats window aggregates

bdougie:
## What users notice

The console's stat strip (Sessions dashboard and Insights) stops paying a full database aggregation on every page view. Any repeat view of the same stats window within a minute — same user navigating between pages, a teammate loading the same dashboard — is served instantly from a server-side cache instead of recomputing the identical numbers.

## Why the cache never hit before

Dashboard clients anchor `since` on their own clock at request time, so two requests for the same logical window ("last 30d") differ by milliseconds. Every request was a window the server had never seen, so no cache at any layer could help, and each one re-ran the whole-window span aggregate (`AggregateSpanStats`: window scan + correlated `EXISTS` per turn row + tool-span count).

## How

- **Snap the window outward to whole-minute boundaries** (since floors, until ceils) before querying. The snapped window always contains the requested one; over-inclusion is ≤60s per edge, which is noise for a whole-window aggregate. Snapping makes identical logical windows collide on one key.
- **Memoize per (org, snapped window) for 60s** — the same staleTime the console already applies client-side, so end-user freshness is unchanged. Errors are never cached. The map is size-capped (1024) with expired-first eviction against pathological key sweeps.
- Swagger description updated and docs regenerated.

No client changes needed; every consumer benefits.

## Testing

- `go test ./api/` — new handler specs: repeated window served from cache (1 driver call), millisecond-unique `since` values collapse to one snapped window, distinct windows cached separately, failures not cached; unit specs for TTL expiry, snap floor/ceil/containment, and key scoping.
- Pre-existing `pkg/spanembed` failure reproduces on clean `origin/main` (environmental, unrelated).

Refs PCC-936 (the deeper fix — rollup tables — stays open; this removes the repeat-recompute cost).

### DIFF
--- a/api/api.go
+++ b/api/api.go
@@ -16,12 +16,13 @@ import (
 
 // Server is the API server for managing and querying the Tapes system
 type Server struct {
-	config    Config
-	driver    storage.Driver
-	logger    *slog.Logger
-	app       *fiber.App
-	metrics   *Metrics
-	mcpServer *mcp.Server
+	config     Config
+	driver     storage.Driver
+	logger     *slog.Logger
+	app        *fiber.App
+	metrics    *Metrics
+	mcpServer  *mcp.Server
+	statsCache *statsCache
 }
 
 // NewServer creates a new API server.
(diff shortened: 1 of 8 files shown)


bdougie:
Closing for now — holding off on this until we decide how to fix stats performance. Branch bdougie/pcc-936-stats-window-cache stays around if we come back to it.

bdougie:
PR was opened prematurely but I'll reopen tomorrow. This is a painful part of my demo watching this take 5-7seconds

PR #258, opened 2026-07-20
✨ feat(api): CORS + short-lived browser read tokens for direct console reads

bdougie:
## What

Tapes-side half of the direct browser→tapes read path (related to PCC-945). The PCC-943 measurement attributed 250–500 ms of every console data read to the Netlify function hop while tapes itself answered in 21–176 ms; this PR gives the browser a way to call the read surface directly.

Three additions, all opt-in and inert by default:

- **CORS on the read surface** — `api.cors_origins` / `TAPES_API_CORS_ORIGINS` / `--cors-origins` (comma-separated origins). Allows `GET,HEAD,OPTIONS` only and the `X-Paper-Auth` header. Since `X-Paper-Auth` is non-safelisted, every browser call preflights, and write methods fail the preflight. Empty config = no CORS middleware at all, exactly today's behavior.
- **`POST /v1/browser-tokens`** — mints a short-lived HMAC-signed token (`tbt_v1.<claims-b64>.<sig-b64>`) bound to the request's org + subject. Meant for trusted server-side callers (the console's server fns) on the same edge-stamped path as every other request — the token only re-asserts an identity that path already carries. Signed with `TAPES_API_BROWSER_TOKEN_SECRET`; TTL via `TAPES_API_BROWSER_TOKEN_TTL` (default 10m). Returns 501 when no secret is configured.
- **`X-Paper-Auth` verification middleware** — a `tbt_v1.` token is verified (constant-time HMAC + expiry) and its signed org claim **overrides** the blind-trusted `X-Tapes-Org-Id` header; its subject overwrites `x-paper-auth-subject`. This closes the org/subject spoofing gap for browser callers. Non-tapes header values (the gateway-bound WorkOS JWT) pass through untouched; invalid/expired tokens get a clean 401 so the console can re-mint or fall back.

## Testing

- Ginkgo specs cover mint/verify round-trip, expiry, wrong-secret, payload-splice tampering, malformed tokens, the middleware's org-override + pass-through + 401 paths, and CORS preflight/actual-response/unconfigured behavior — all driven through `server.app.Test()` with the full middleware chain.
- `go build ./...`, `go vet ./...` clean; `api` + `pkg/config` suites green. The `spanembed` / `storage/postgres` failures are pre-existing on `main` (they need the Dagger Postgres environment).
- Swagger/OpenAPI regenerated (`make swag && make openapi`).

## Rollout

Inert until deployment sets `TAPES_API_BROWSER_TOKEN_SECRET` and `TAPES_API_CORS_ORIGINS=https://console.papercompute.com`. The gateway must also pass `OPTIONS` preflights and `tbt_v1.` `X-Paper-Auth` values through to tapes. Console counterpart: papercomputeco/console.papercompute.com PR from `pcc-945/console`.

related to PCC-945


bdougie:
@greptileai review

PR #267, opened 2026-07-27
✨ feat: ingest fidelity, published contracts, and the first fixture-grade wire recordings

jasonwc:
The raw layer is meant to be the source of truth, with the projection rebuildable from it at any time — a bug should be a re-run away from repair, not a data-loss event. Three things quietly failed that: the raw layer was not actually raw, a failed reduction was unrecoverable, and the write surface every capture path targets had no published contract.

- **`raw_response`** stores the upstream body verbatim beside the reduced one. Kept encoded rather than decompressed — re-compression is not byte-identical. A separate drop marker distinguishes "never had bytes" from "had them and hit the cap"; without it a gap reads as an absence.
- **Recovery**: the derive read now fetches those bytes for exactly the turns whose reduction is unusable, and reduces before handing the row over. Healthy turns fetch nothing.
- **Change feed**: `content_hash`, a cursor, and a provenance tier on the span projection, computed at write time.
- **Contracts**: an OpenAPI document for the write surface, the previously-undocumented skills and derive-run endpoints, and a gate that fails when a served route is undocumented or a documented route is unserved.
- **Recordings**: the first 17 fixture-grade wire bundles, plus an executable scrub gate.
- Deletes the unused `/v1/ingest/batch`; makes the DB-backed suites runnable locally.

### Worth a reviewer's attention

**The cursor is a transaction id, not a sequence.** Any cursor assigned inside a transaction orders writes rather than commits, so concurrent passes can commit out of order and a `> cursor` poll silently skips one. Postgres will say which *transactions* are still in flight and will not say that about a sequence, so the xid is what makes a safe read expressible at all. The bounded read ships as `ListChangedSpanTurns` / `ListChangedSpans` rather than as advice, because a consumer writing the poll by hand writes the lossy one. The test commits the lower cursor last — the only arrangement that exposes this.

**Reduction-on-read lives in the storage layer, not `pkg/derive`.** The deriver stays a pure function of the rows it is handed; reduction depends on provider, content type and encoding, none of which are projection facts. The derive path itself is unchanged — it just stops being handed an empty reduction.

**`content_hash` is length-prefixed per field**, or `("ab","c")` collides with `("a","bc")` and the feed reports "unchanged" for a real edit. `session_id` is excluded because the upsert `COALESCE`s it, so hashing the supplied value would report a change the table did not make.

**The migration builds no index, deliberately.** Migrations run at process start under an advisory lock every serve path waits on, so `CREATE INDEX` blocks the table for the whole build and `CONCURRENTLY` merely trades that for a longer wait every starting pod blocks on. The operator commands are recorded in the migration for when a consumer exists.

**The recordings are real captures in a public repo.** Identifying values carry synthetic stand-ins, same-length in request bodies so `content-length` stays valid. Session ids are kept deliberately: the envelope and sub-thread attribution the corpus exercises are keyed on them. The scrub bar is an executable gate matching identifier *forms* rather than known values, because a first pass of human review over these same bundles caught the credentials and missed a device fingerprint nested inside a JSON-encoded field.

Related to PCC-982, related to PCC-1003, fixes PCC-1028.


jasonwc:
@greptile-apps review

jasonwc:
@greptile-apps review

PR #272, opened 2026-07-29
✨ feat: OpenAPI v3 implementation

jpmcb:
# 👨🏼

Implements a `pkg/tapesoapi/` package and replaces goswag with OpenAPI native implementation.

Closes PCC-1047

# 🤖


jasonwc:
Merged `main` into this branch (`50e2684`) to clear the conflicts — resolution notes for review:

- **Goswag artifacts stay deleted** (`api/openapi.yaml`, `docs/docs.go`, `docs/swagger.*`): main had regenerated them for the new admin endpoint, but under compiled contracts there is no stored artifact.
- **The real port**: main's raw-turn attribution-repair endpoint (#-overlay work) is re-expressed natively — registered through the route table with its full contract (request schema from `storage.RawTurnAttributionRepairRequest`, all seven response codes including the 202 projections-pending case), and the handler's goswag `@` block converted to prose pointing at the registration. The coverage gate accepted it on first run — the exact staleness-proofing this PR introduces, working as designed.
- `CONTRIBUTING.md`: this branch's toolchain list wins (no `swag`), keeping the `TEST_POSTGRES_DSN` note.
- `api/api.go`: route-table structure wins wholesale.

Verified with the Dagger-provisioned suite (`make test`): **58 packages ok, zero failures**, coverage + aggregation gates green.

Remaining for merge: review (Greptile never ran on this PR — likely needs a manual trigger) and a post-merge consumer follow-up: paper's `paper-api-tapes/vendor/openapi.yaml` sync procedure switches from the checked-in spec to `tapes dev openapi api` output.

jasonwc:
@greptile-apps review

jasonwc:
Greptile's two findings triaged at `1a55377` — **both confirmed in code and fixed**, and both sat squarely in what a generated client would consume:

1. **Nested `servers` overrides survived republication** (`RewritePrefix` stripped root-level only). A cassette declaring a path-item or operation-level `servers` would have generated clients calling its private listener, bypassing the proxy — and capture — entirely. Now stripped at every level; pinned by a spec asserting no `cassette.internal` survives while the operation content does.
2. **Reference-typed example values were never walked** during component namespacing — `Components.Examples` and media-type `examples` values can be Reference Objects, and their targets got renamed out from under them (dangling at best, cross-cassette collision at worst). Now rewritten in both places; pinned end-to-end through `Compile`.

Both specs verified to fail on the pre-fix code. `pkg/tapesoapi` + `api` suites green; the compile pipeline's own lint gate even caught a missing `operationId` in my first test fixture, which was satisfying to watch.

jasonwc:
Re-review triage at `ac01da6`:

- **"Webhook server overrides survive" — new, valid, fixed**: `RewritePrefix` stripped servers at every level under `paths` but copied the root `webhooks` value (OpenAPI 3.1) unchanged. Webhook path items now get the same treatment — nothing in a republished document may name the cassette's own listener. Pinned by a spec verified to fail on the pass-through; webhook content survives the strip.
- The other two threads (nested path/operation servers; example references) are **re-anchors of findings already fixed** in `1a55377` — `stripNestedServers` and `walkExampleRef` are in the reviewed code at those very lines.

Lint clean (`golangci-lint` 0 issues), 100 tapesoapi specs green.

PR #276, opened 2026-07-31
🔧 fix(api): stop letting callers assert their own tenant on reads

jasonwc:
tapes no longer has a concept of an organization.

Previously, reads accepted an `X-Tapes-Org-Id` header that scoped queries to whatever org the caller named, and writes stored whatever `org_id` the payload carried. Nothing verified either — the header was bookkeeping that looked like isolation. A tapes instance serves one tenant; if you need multi-tenancy, run one instance per tenant and enforce identity at your own perimeter (a reverse proxy, gateway policy, or network boundary), where it can actually be checked.

**What changed:**

- The `X-Tapes-Org-Id` request header and the `tapes search --org` flag are removed. Reads always operate on the instance's single tenant.
- Ingest ignores any `org_id` supplied in payloads (both the wire and transcript lanes) — the field remains accepted in the wire contract, so existing clients work unchanged, but it no longer affects where rows land.
- Transcript dedup no longer keys on org: the same content is the same row, and nothing a request carries can create a duplicate.
- `auth_subject` (who wrote a session) is unchanged.
- A migration verifies at upgrade time that all existing rows already carry the single default org value, and halts with a clear error if any don't — so no data can silently become unreachable through the API. (In practice nothing ever wrote another value; the check makes that a guarantee instead of an assumption.)

The internal `org_id` columns still exist and are pinned to a single constant at the handler boundary; removing them from the schema is a follow-up.


jasonwc:
@greptileai review

jasonwc:
@greptileai review

jasonwc:
@greptileai review

PR #280, opened 2026-07-31
🔧 fix(ingest): keep capture-side duration and timestamp on raw-only rows

jasonwc:
Raw-only payloads carry verbatim upstream bytes and no client-side reduction, so ingest reduces them server-side. The reducers see only those bytes, which carry neither the call's duration nor the instant it was captured — two capture-side facts that previously reached the stored row only because the producer reduced live. This closes both gaps.

**Duration.** `Usage.TotalDurationNs` was never stamped on the raw-only path, landing derived spans with a NULL `duration_ns`. Ingest now stamps it from `meta.elapsed_seconds` during server-side reduction, matching what a live-reducing producer would have recorded.

**CreatedAt.** Under raw-only it silently changed meaning from capture time to ingest time, because the reducers stamp `time.Now()` and that clock now runs on the server. It now means "when the turn happened, never when the server heard about it", resolved from the envelope in precision order: `meta.captured_at` (new optional field, exact) → `meta.ts_request` + `meta.elapsed_seconds` → `meta.ts_request` alone → whatever the reducer produced. Reusing `ts_request` keeps `CreatedAt` on the same clock the deriver already uses for span start times, and correctly dates backfilled recordings with no producer change.

Producers that don't yet send a timestamp fall back to the reducer's value; the fallback rate is observable per provider and field via the new `tapes_ingest_rawonly_stamp_total{provider,field,source}` counter rather than being silent. Providers whose wire format carries its own creation time (e.g. OpenAI Responses `created_at`) are preserved untouched.

Dual-send payloads are unaffected — the stamps apply only where the server is the reducer.

related to PCC-1044

### DIFF
--- a/ingest/elapsed_guard_test.go
+++ b/ingest/elapsed_guard_test.go
@@ -1,0 +1,33 @@
+package ingest
+
+import (
+	"math"
+	"testing"
+)
+
+// usableElapsed guards every conversion of meta.elapsed_seconds into
+// timing data; a value it wrongly accepts becomes a bogus duration or a
+// CreatedAt far in the future.
+func TestUsableElapsed(t *testing.T) {
+	cases := []struct {
+		name    string
+		elapsed float64
+		want    bool
+	}{
+		{"typical call", 1.75, true},
+		{"upper bound inclusive", maxElapsedSeconds, true},
+		{"zero is absent", 0, false},
+		{"negative is corrupt", -3, false},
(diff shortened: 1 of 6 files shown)


jasonwc:
@greptileai review

jasonwc:
@greptileai review

jasonwc:
@greptileai review

PR #282, opened 2026-07-31
✨ feat(storage): add local SQLite core mode (PCC-XXX)

jpmcb:
# 👨🏼

# 🤖

---

Implements a Docker-free, single-process SQLite core mode for `tapes serve` and `tapes start`. SQLite is now the default at `.tapes/core.sqlite`; `--postgres` retains the existing multi-process PostgreSQL deployment.

- Adds raw-turn capture, derive queue, session/span projection, API reads, stats, export, and transcript reconciliation through `pkg/storage/sqlitecore`.
- Uses WAL, foreign keys, full synchronous writes, serialized writes, and an exclusive lifetime file lock.
- Hardens local derivation against deleted/pre-ingest sessions, capture ordering, duplicate identities, dangling links, zero completion timestamps, raw-response reduction failures, and malformed synthetic roots.
- Adds SQLite schema-version tracking, deletion subtree/queue cleanup, and serializes all SQLite writes.
- Keeps vector search, embeddings, skills, and standalone services PostgreSQL-only. `tapes start --postgres` retains the existing separate derive-worker topology.
- Note: envelope-less proxy captures now create the matching synthetic session on both SQLite and PostgreSQL, closing the capture-to-derive gap.

Validated with `make format` and `make test`.


jpmcb:
Closing as this was primarily an experiment to understand the surface area of how a local sqlite compares to postgres

PR #283, opened 2026-07-31
🧹 chore(cli): make tapes the server surface and tapesctl the client

jasonwc:
`tapes` shipped as one binary that both ran the services and acted as their client. The client half now lives in [`tapesctl`](https://github.com/papercomputeco/tapesctl), so keeping it here means two implementations of the same reads drifting against a single contract.

This removes every command whose job was to talk to the system as a user:

| Removed | Replacement |
|---|---|
| `tapes start` | `tapesctl start` |
| `tapes export` | `tapesctl export` |
| `tapes seed` | `tapesctl seed` |
| `tapes sessions` | `tapesctl sessions list` |
| `tapes skill sync` | `tapesctl skill sync` |
| `tapes search` | `tapesctl search` |
| `tapes skill generate\|list` | `tapesctl skill generate\|list` |
| `tapes deck` | *none — the console owns the ROI/session-browsing surface* |

`pkg/start` goes with them — it was the daemon state/lock/log manager for `tapes start` and had no other consumer. So does `cmd/tapes/inprocessapi`, whose only consumer was `skill generate`. `pkg/seed` stays; the API's admin seed endpoint still uses it. Span search itself stays server-side — the `/v1/search/spans` API and the embed worker are unchanged; only the client command moved.

What remains is what needs the database, the process, or the local filesystem: `serve` and its workers, `local`, `dev`, `init`, `config`, `auth`, `status`, `backfill`, `version`.

One behavior is lost with no direct replacement: `tapes sessions --postgres`, which listed sessions without a running server. `tapes serve api` plus `tapesctl sessions list` covers the same ground.

Docs are updated throughout: quickstarts in the README and the book now install both binaries and use `tapesctl` for capture and reads; the deck page is gone.

related to PCC-986


jasonwc:
@greptileai review

PR #288, opened 2026-08-04
✨ feat(extproc): fold the ext_proc capture adapter into tapes

jasonwc:
related to PCC-988

extproc moves into this repo as its own deployable — the Envoy `ext_proc` capture adapter, imported verbatim from tapes-extproc at 1768c23 (v0.13.0) and integrated with **zero behavior change**. It stays a separate binary/runtime unit from the API server: the Envoy hot path must not depend on core's DB or derive-worker health.

**Layout** follows this repo's own conventions (service packages top-level, `package main` under `cli/`): `extproc/` as a peer of `ingest/`, `extproc/headers/` nested under its only consumer, binaries at `cli/tapes-extproc` and `cli/extproc-replay-server`.

**Commits tell the story** (review in order; the import commit deliberately doesn't compile alone — the PR squash erases the intermediate states):
1. Verbatim tree import — paths changed, bytes not.
2. Package/import rewrite into this module; deps pinned to extproc's tested versions.
3. Cross-repo constant mirrors collapse into direct `ingest.MaxRawResponseBytes` / `ingest.MaxIngestBodyBytes` references; the two now-vacuous pinning tests are deleted.
4. Both vendored envelope corpora + their sync scripts retire — tests read `fixtures/` in place (closes the Go-side of the PCC-1002 sync).
5. Build/test/ship from here: a dedicated DB-free Dagger check (`CheckExtproc`) with its own PR job — deliberately not folded into the unit `Test` target, which CI does not run — plus `Dockerfile.extproc` and an image-build job.

**Deliberately unchanged:** the raw-lane interlock (`ingestCanDecodeEncoding`) still claims ingest decodes only gzip, even though #287 gave ingest zstd + stacked decoding. Widening it is the PCC-1044 extproc follow-up and belongs in its own PR; comments now describe the predicate as a conservative floor rather than a symmetry.

**Costs worth knowing:** importing `ingest` for two constants grows the sidecar binary ~5.5 MB (compile-time only — Fiber et al. link in but nothing couples at runtime); `klauspost/compress` moves 1.18.4 → 1.19.0 module-wide (the version extproc's zstd path was tested on); the two comment-only edits to `ingest/ingest.go` and `proxy/header/header.go` fix now-stale cross-repo pointers.

Gates: full Dagger suite 57 packages green (extproc's three included), `make test-extproc`, `make parity`, extproc image builds, lint clean.


jasonwc:
@greptileai review

PR #292, opened 2026-08-05
✨ feat: seal and publish the OpenAPI contracts

jasonwc:
related to PCC-1065

Stage S1 of the client-generation track: seal the published OpenAPI contracts so generated clients have a stable, reviewable upstream.

- `api/CONTRACT` + `ingest/CONTRACT` record the **prose-stripped** `Fingerprint()` (the exact document a deployed binary serves — doc-comment prose requires a source tree and is excluded; Go-value prose is always present). Ginkgo gates in both packages fail with copy-paste bump instructions when the published shape moves; a third spec per package pins that CONTRACT does *not* hold the prose-included fingerprint, which is the plausible wrong paste.
- Seal bumps must ride the change that moved the contract — the seal is the record that a reviewer was told the surface changed.
- `make contracts` (via a new dagger `contracts` function) writes both prose-included YAMLs to `build/contracts/`; output proven byte-identical between macOS-local and linux-container compiles.
- Release workflow attaches `tapes-api-<tag>.yaml` / `tapes-ingest-<tag>.yaml` as assets — the CORE compiled documents, not the runtime aggregate (a running server's `/openapi` merges mounted cassette specs, which is a property of a deployment).

Gates: build/vet/lint/full test suite/parity green; tamper run proves the seal fails.

### DIFF
--- a/.dagger/contracts.go
+++ b/.dagger/contracts.go
@@ -1,0 +1,38 @@
+package main
+
+import (
+	"dagger/tapes/internal/dagger"
+)
+
+// Contracts compiles both published OpenAPI documents and returns them as a
+// directory: tapes-api.yaml and tapes-ingest.yaml.
+//
+// These are the documents consumers vendor. Nothing in this repository reads
+// them, and they are not checked in — a contract file is a copy of what the
+// server already states, and a copy can go stale. What is checked in is the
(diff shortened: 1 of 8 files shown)


jasonwc:
@greptileai review

PR #295, opened 2026-08-06
:sparkles: feat(fixtures): pin the content-encoding decode policy to a shared corpus

jasonwc:
Capture policy has two implementations — the Go one here, and a Rust one in the client.
Their agreement has been a point-in-time claim by whoever last read both files, and that
decays silently. It decayed: the client dropped every compressed request body while this
path decoded it fine, and nothing failed until someone ran a session and noticed the rows
were missing.

This adds a shared corpus for content-encoding decode, so the two implementations are
gated against the same cases instead of against each other's source. It generalises the
pattern already used for the envelope fixtures.

## Format

Cases specify a body either as a **recipe** — a plaintext, a list of encodings to apply,
an optional truncation — or as literal base64 where the bytes *are* the assertion.

Recipes are the default because compressed output is not stable across implementations:
Go's `compress/gzip` and Rust's `flate2` do not emit identical bytes, and neither do the
two zstd libraries. Pinning compressed bytes would assert compressor identity, which is
not the policy and which no second implementation could satisfy. The cap cases decode to
32 MiB, which is not committable either way.

## 25 cases

identity (5) · supported codings (4) · stacked layers (2) · salvage (4) · limits (4) ·
errors (6)

## Three questions the corpus forced into the open

**zstd bounding.** Two 14-byte frames differing only in the window descriptor show the
window bound is observable from outside the decoder, and both implementations place it at
log2(32 MiB). That agreement is now a gate. Maximum-memory is asserted nowhere — with the
window already at the output cap, no constructible frame distinguishes a decoder that sets
it, so a case would assert a library's API surface rather than the policy.

**Empty bodies.** The case pins the decoder. "Don't call the decoder for a bodiless
request" is recorded as a caller precondition that no decoder test asserts, which makes an
early return in a caller conformant rather than divergent.

**Failure classes.** Three — unsupported, oversize, undecodable — because both
implementations already distinguish exactly three. Corrupt input and a refused window stay
merged, with the difference in an explicitly unasserted field. Splitting them is a change
to both implementations first and the corpus second.

## It found something immediately

`DecodeContentEncoding([]byte{}, "gzip")` errors while `([]byte{}, "zstd")` returns success
with zero bytes — identical input, opposite outcomes, decided by which decoder the header
named. That is documented here as a case rather than quietly fixed, and repaired in the
follow-up.

## Gates

An oracle over every case, a digest seal so a stale vendored copy fails, and coverage rules
so deleting the last case for a behaviour fails loudly. All three are mutation-tested.

Two failures on a full `go test ./...` reproduce on a clean tree and are unrelated: the
OpenAPI contract seal, and the Postgres suite, which needs the database `make test`
provisions.

related to PCC-1126


jasonwc:
@greptileai review

Both findings addressed.

**Salvage case portability.** The truncation is now an absolute `keep_head_bytes: 9` instead of a ratio. 9 comes from the frame structure — 4-byte magic + the smallest legal frame header (2) + a block header (3) — so byte 10 is the earliest position at which any encoder's first byte of block *content* can appear, and a block that has delivered no content cannot have produced output whatever its type. A larger frame header only pushes that further out, so the bound holds for every encoder rather than for the one that generated the fixture. The schema had no absolute-head form, so `keep_head_bytes` was added alongside the existing two, and the README now says what each form holds fixed across compressors. The asserted outcome is unchanged: probed at keeps of 4/5/6/7/8/9/10/12/44 bytes, the decoder returns unexpected-EOF with zero output at every one.

**Coverage predicates.** The gate is now a declared table of {rule, the case name that pins it, predicate}, with the predicate run against *that named case* rather than searched across the corpus. All three failure modes verified: deleting a pinning case, editing it so it no longer pins its rule, and adding a case no rule names — the last closes the table in the other direction so it cannot fall behind the corpus.

Worth recording on your DIGEST question: the seal alone would **not** have caught a deletion. It is recomputed from whatever cases exist, so delete-and-reseal is a legal two-line green diff. It is the named-case gate, not the seal, that stops that.

jasonwc:
@greptileai review

One more case pair on this branch, from a defect found reviewing the client that consumes this corpus.

Its gzip binding stopped after the first member of a concatenated stream and reported success, while `compress/gzip` here spans members by default. Same input, different captured body, no error on either side — exactly the drift this corpus exists to make loud, and invisible until the two were run against the same case. `gzip-concatenated-members` pins it.

`zstd-concatenated-frames` pins the case that already agrees. The two codings go through unrelated libraries on each side, so agreement on gzip says nothing about zstd, and a rule that is only pinned for one coding is one a future binding swap can silently drop for the other.

The body spec gained a `members` form for this. It splits the **plaintext** before compressing, not the encoded stream — a split of encoded bytes would cut at an offset only one encoder's output has, and the case could no longer assert against the original plaintext.

PR #296, opened 2026-08-06
:sparkles: feat(fixtures): specify the drop-reason taxonomy and its boundary

jasonwc:
**Stacked on #295** — review that first; this branch contains its commits.

## One empty-body rule for every content-coding

#295 documented, rather than hid, an inconsistency: an empty gzip body errored while an
empty zstd body returned success with zero bytes. Identical input, opposite outcomes,
decided only by which decoder the header named. An empty body under a claimed coding means
something went wrong — a body lost in flight, a truncated read — and one of the two codings
was swallowing it.

`refuseEmpty` now runs in both the gzip and zstd branches. It sits **inside each supported
branch rather than once above the switch**, deliberately: an unsupported coding must stay
*unsupported* whatever the bytes are, since nothing is read and the body cannot change the
answer. A guard above the switch would report an unknown coding plus an empty body as
*undecodable*, collapsing two failure classes that are kept apart on purpose.

One downstream consequence worth stating: a POST to a turn path carrying a content-encoding
header and no body now drops as `request_decode` rather than dispatching an empty request.
That is the outcome gzip already produced for the same input. Health probes are unaffected —
the non-turn-request check runs first.

## Specify the drop-reason taxonomy

The ext_proc adapter carries a vocabulary of drop reasons that existed in Go only and was
written down nowhere. A second implementation had no way to know which of them it was
supposed to reproduce.

They split 7 / 7, by a sharper test than intuition: *could a conformant implementation with
different plumbing produce this reason at all?*

**Contract** — `upstream_status`, `non_turn_request`, `request_decode`, `empty_response`,
`unknown_provider`, `response_decode`, `reducer_error`.
**Transport** — `client_disconnect`, `upstream_no_response`, `missing_status`, `sem_full`,
`ingest_reject`, `ingest_timeout`, `marshal_error`.

The close calls are argued in their case files. Two worth surfacing:

- **`empty_response` and `upstream_no_response`** both mean "no response bytes" and they
  split. A response phase that completed carrying nothing is a property of the exchange; a
  stream torn down before its response phase is a property of the connection, and only a
  streaming interception path can tell them apart.
- **`missing_status`** reads like contract and is classified transport. It is reachable
  only because ext_proc can deliver a body before its headers; a path that reads a status
  off a response object always has one, so it can neither conform nor diverge.

The contract half now lives in `pkg/capture`, and the adapter's constants are conversions
of it, so this build cannot change a wire-visible label on its own. Conformance is gated in
both directions: a reason in code but not the corpus fails, and so does the reverse.

## Two things recorded rather than changed

`DefaultLargeTurnThreshold` is **not** contract. It labels one metric; two builds that
disagree about it capture, dispatch and store byte-identical turns and differ only in a
sizing dashboard. A threshold that dropped or truncated a turn would be contract.

The client currently has no upstream-status gate and no turn-path gate, so a failed
exchange, or a GET carrying a JSON body, becomes a stored turn there and nothing here.
Both are recorded as known divergences. Neither would have been found by capturing
sessions and comparing — they produce extra rows, not missing ones.

related to PCC-1126


jasonwc:
@greptileai review

The caller-precondition comment is corrected. Verified your reading is right: the gateway adapter calls the decode unconditionally at request end-of-stream with no check that bytes arrived, so the claim that both paths return early for bodiless requests was false.

The comment now says which path returns early and which does not, notes that ordinary bodiless traffic is filtered by the turn-request gate that runs *before* the decode error is consulted, and states plainly what is left over and intended: a POST to a turn path claiming a coding with no body terminates as a `request_decode` drop. Documentation only — no behavior change.

Five lint findings this branch introduced are also fixed (a `goconst` from the corpus test spelling two strings once per example, a `gofumpt` wrap, and three hand-rolled searches replaced with `slices` calls).

jasonwc:
The `tapes:check-lint` red is not this branch.

I ran that check locally through the same Dagger engine version the module pins (v0.21.7 — a checksum-verified release binary, since the engine on this machine was older than the module requires and refuses to load it). It passes:

```
✔ .checkLint: String!   1m12s
0 issues.
```

That is the actual check, not an approximation of it. Alongside it, `golangci-lint v2.8.0` against the repo's own config reports `0 issues.` directly.

Every failing status here reports `error` rather than `failure` — what a runner that did not complete looks like, as opposed to a check that ran and found something. With the same check passing locally, the remaining difference is the hosted runner.

PR #308, opened 2026-08-13
📚 docs: flatten the docs to one Markdown file per page, and correct them

jasonwc:
> Heads-up @matt: docs + build-plumbing only (docs/, flake.nix's mdbook, the docs make targets, one AGENTS.md line) — nothing under api/, cmd/, or the cassette surfaces. Should not collide with the cassette track.

The docs move to one flat Markdown file per page with `title`/`description` frontmatter, and every page was corrected against the code while it was being touched.

**The load-bearing fix:** every capture example pointed `tapesctl start` at :8081 — the read API. `POST /v1/ingest` is registered only on the ingest server (:8082), so the documented quickstart captured nothing while printing success. All 17 start/capture/sync examples now use 8082; the surviving 8081 mentions are prose about the read API.

The rest, per page: the contract-existence error in apis.md/development.md corrected (the sealed contract exists and contributors will meet it); Deck removed everywhere including the dead link and the duplicated "browse interactively" section (rewritten around `--api-web-ui`); integrations.md rewritten to the real harness lanes (claude/codex/pi via start, codex-app via capture, plugin-install-before-pi, sync) with the OpenCode-start and OpenClaw sections removed; the cassette walkthrough teaches `tapesctl cassettes <name> <method>`; configuration.md now distinguishes the server's config resolution from the client's; phantom flags and duplicate lines out; telemetry.md states plainly that the client has none. mdBook scaffolding (book.toml, SUMMARY.md, build-cloudflare.sh) is deleted — the deployed book keeps serving until the new docs site is live; nothing here touches that deploy.

## Verification

15/15 pages pass a dependency-free frontmatter check (negative-controlled); 23 relative links resolve; **103 command examples verified** against a server built from this tree (v0.35.0-2) and tapesctl v0.4.0 — the four cassette examples ran live against a discovery stub with character-identical output; `nix flake check` green; forbidden-terms grep (server-CLI verbs, Deck, 8081-for-ingest, the retired cassette spelling) all zero.

related to PCC-1182


jasonwc:
On the red **Cloudflare Pages** status: expected, and it stays through the interim. This PR deletes the mdBook scaffolding (`book.toml`, `build-cloudflare.sh`), so the docs.tapes.dev Pages project can no longer build from this branch — or from main once this merges. That is the designed interim state: **Pages keeps serving its last good deploy**, so docs.tapes.dev continues to serve the current book unchanged until the new docs site goes live and the Pages project is retired with a 301 (the final step of the migration, an account-side action). The status will read red on main until then; nothing user-facing changes. The check is not a required context.

PR #310, opened 2026-08-13
✨ feat: /v1/stats accepts an auth_subject filter

jasonwc:
Adds an optional `auth_subject` query parameter to `GET /v1/stats`, filtering every aggregate to sessions ingested under that gateway-verified subject. Absent or empty → unchanged org-wide behavior; present → exact match; no match → zeros, not an error.

Why `auth_subject` and not `subject`: the sessions list API already filters on `auth_subject`, and existing consumers already append `auth_subject=` to stats URLs expecting it to take effect — one vocabulary for one concept.

- All eight aggregates honor the filter, including `tool_calls`, which lives in a separate subquery with no session join — handled with an uncorrelated semi-join (one hashed subplan over the existing `(org_id, auth_subject, id)` index) rather than a per-span correlated probe.
- Deliberate semantic edge: filtering tightens the trace→session LEFT JOIN to an inner match — an unattributed trace has no subject, so it counts for nobody under a filter; unfiltered behavior is byte-identical.
- Whitespace subjects are preserved, not trimmed to NULL: a blank-looking subject means "nobody by that name," never a silent widening to the org.
- Contract seal bumped in the same commit; sqlc output matches the pinned generator; 12 new specs; mutation-testing the `tool_calls` clause reproduces the org-wide leak this exists to prevent.

Heads-up @yeazelm  diff is confined to the `/v1/stats` surface (9 files); the one shared edit is the `SpanStatsReader` interface signature (postgres driver + test stub). No release cut — riding the next scheduled one.

Note for CI: the DB suite ran locally against `pgvector/pgvector:pg17` (the pinned ECR Public image 403'd locally); `pkg/spanembed` has a pre-existing first-run flake that reproduces on unmodified main.

related to PCC-1192

### DIFF
--- a/api/CONTRACT
+++ b/api/CONTRACT
@@ -13,4 +13,4 @@
 #
 # api/openapi_seal_test.go recompiles and compares. If it fails, it prints the
 # value to write here. Bump it in the same change that moved the contract.
-sha256:26755fe576fbd4a9eb57639472a00e8ca1f9de7be1e27296bd85975add46e9c8
+sha256:2a36f0b39cca676a7b877bcde488967c227b3ece431bc84d2c3329a8a688a97e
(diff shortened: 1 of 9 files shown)


jasonwc:
@greptile review

PR #314, opened 2026-08-14
📚 docs: the README points at the docs home that exists

jasonwc:
The Documentation badge and the agent-guides link pointed at `/guides/`, which now 301s — both move to `https://tapes.dev/docs/`, verified 200 live. Also removes two dead entries from the dagger source ignore list that referenced the site directory deleted with the docs flatten (module loads clean without them).

related to PCC-1186


jasonwc:
@greptile review — initial review

PR #337, opened 2026-08-28
✨ feat: arm filter claims only against readable published views

yeazelm:
🤖

## Summary
- Core previously admitted a manifest filter claim without ever checking it could read the claimed published view; a deployment where the view's grants had not landed yet served an opaque 500 on every request using the claimed param.
- Claims are now **armed** separately from admission: a probe (`SELECT primitive_type, primitive_id, <claim column> FROM <view> WHERE FALSE`) verifies existence, readability, and the declared shape with core's own database role before the claim takes effect.
- An admitted-but-un-armed claim degrades to the unclaimed semantic — the param is ignored byte-identically — and is reported as a first-class rejection in `GET /v1/cassettes` `problems[]` naming the view and the SQLSTATE. Armed-then-broken at request time remains a loud 500 (the never-silently-unfiltered contract is unchanged).
- Arming self-heals: the probe runs once per refresh pass (outside the manifest 304 short-circuit, which never re-runs admission), so a grant landing later arms the claim within one refresh interval with no restart. Arm/disarm transitions fire the registry-changed hooks.

## How it works
Validation stays grammar-only and admission stays no-I/O. The probe is injected as a nil-safe func on the runner config (same seam as `ReservedParams`), closing over the storage driver's new `PublishedViewProber` capability at server construction — the runner package stays storage-free, and a nil probe arms everything (non-probing drivers keep today's behavior). The armed set lives on the registry; `ClaimsFor` returns armed claims only, so the request path needed no changes. Rejections file under a stable claim-qualified subject and clear on arming; a pre-existing subject-key asymmetry in source-rejection clearing is fixed alongside.

## Behavior matrix
| State | Before | After |
|---|---|---|
| View readable | filtered | filtered (armed) |
| View missing/unreadable at admission | 500 on every claimed request | param ignored byte-identically + rejection in `problems[]`; arms automatically when readable |
| View breaks after arming | 500 | 500 (unchanged, deliberate) |

## Test plan
- [x] `[postgres]` probe specs: readable → nil; missing view (42P01), missing column (42703), revoked SELECT (42501) → error
- [x] Runner specs: probe-fail → un-armed + rejection; self-heal across a 304-fresh refresh; disarm notifies hooks on transitions only; withdrawal requires fresh probe on re-admission; nil probe arms everything; first-claim-wins holds
- [x] Handler spec: un-armed claim → byte-identical unfiltered response with no storage contact; existing claimed-but-broken→500 contract test passes unchanged
- [x] Discovery spec: un-armed claim appears in `problems[]`, clears on arming
- [x] Full Dagger suite green (141/141 postgres specs); race detector clean on the runner; docs updated in the same change

Related to PCC-1280


yeazelm:
🤖 The finding was valid and is fixed (branch also rebased onto main, which now carries #336):

**Transient failures no longer move arming state.** Probe outcomes are classified where the driver types live: the database answered exactly when a `PgError` is in the chain (any SQLSTATE) — only that is a definitive verdict. Everything else (context deadline, connection refused, closed pool) is wrapped as a transient probe error and flows up through layer-owned types, so the runner stays storage-free. On a transient failure the runner retains prior state: an armed claim stays armed, an un-armed claim keeps its recorded definitive rejection text untouched, and nothing arms without a definitive success — with a Warn log and no rejection/hook noise. Rationale in the code and docs: if the database is unreachable, list queries fail loudly on their own; a blip must never silently flip filtering semantics.

New coverage: runner specs for retain-across-transient (armed, un-armed-with-rejection, and never-armed cases), adapter specs for the transient/definitive translation (grammar failures stay definitive), and a storage spec proving an expired context classifies transient while a real SQLSTATE does not.

Full Dagger suite green post-rebase (52 packages, 0 failures); lint 0 issues on touched code.

@greptile-apps review

PR #338, opened 2026-08-31
✨ feat: advertise core trace and span primitives, stamp the session on standalone trace reads

yeazelm:
🤖

## Summary
- Core's aggregated entity registry now advertises `trace` and `span` beside `session` — the derived read model's primitives, declared in the same shape cassettes use. Their ids are deterministic content-derived strings that survive re-derivation, so annotating consumers can key rows off them; a span's canonical external id is `<trace_id>/<span_id>` (composite within a trace). Advisory as ever — nothing gates on advertisement.
- `GET /v1/traces/:trace_id` now stamps `session_id` — the standalone lookup is the one surface whose caller does not already know the owning session and needs it to navigate. The session-scoped composite deliberately leaves the field empty (`omitempty`), so those payloads and the generated fixtures stay byte-identical, preserving the "ids belong to the session envelope" doctrine everywhere it applies.
- Contract reseal rides the same commit as the surface change, per the seal's own rule.

## Test plan
- [x] Entity aggregation specs updated: core-native set is session/trace/span, cassette entities compose after them, withdrawal/replacement semantics unchanged
- [x] Discovery spec pins the new registry contents + ETag movement
- [x] New standalone-trace spec: `session_id` present on `GET /v1/traces/:trace_id`; `BuildTraceDetail` neutrality spec proves the composite (and fixtures) carry no duplicate
- [x] Full Dagger suite green (exit 0) at this exact SHA; gofmt/lint clean

Related to PCC-1280


yeazelm:
🤖 Both items addressed in 796f502:

**CI lint** — a `usestdlibvars` finding in the new standalone-trace spec (string method literal → `http.MethodGet`); reproduced and re-verified with the canonical Dagger lint (exit 0).

**Modeling finding (valid)** — the standalone lookup now returns a dedicated `StandaloneTraceDetail` type: `session_id` is a required field there, and `TraceDetail` sheds the optional field entirely — generated clients see the guarantee where it exists, and the session-scoped composite schema no longer advertises a field its payloads never emit. Contract resealed in the same change; the composite-neutrality spec now asserts the serialized composite carries no `session_id` at all. Full Dagger suite green at this SHA.

@greptile-apps review

yeazelm:
🤖 Valid — fixed in 2182f66: `session_id` on the standalone response now carries the reflector's explicit `oas:"required"` annotation, so generated clients model the guarantee as required. Contract resealed in the same change; canonical lint and full Dagger suite green at this SHA.

@greptile-apps review

PR #339, opened 2026-09-01
🧹 chore: document the span external-id separator

yeazelm:
🤖

## Summary
- Prose-only: the advertised span composite external id joins with `~`, not `/` — both components embed provider-verbatim ids (full of `_`/`-`), and an escaped slash does not survive gateway escaped-slash normalization (empirically: envoy 307s `%2F` to the unescaped path). The merged comment documented the dead convention; this aligns it with what consumers actually mint.
- The sealed contract is untouched (doc comments are outside the fingerprint).

## Test plan
- [x] cassetterunner suite green; gofmt clean; no schema/route change by construction

Related to PCC-1280


yeazelm:
🤖 Valid as a documentation gap — closed by stating the actual contract rather than pretending the charset is excluded: the composite id is primarily an opaque round-trip handle (attachment-style consumers store and replay it whole, never reconstruct), and splitting on the first `~` is declared a display/navigation convenience, best-effort by contract. Provider-verbatim components are not formally barred from `~`, and no unreserved-and-excluded character exists once provider ids are embedded — so opacity, not escaping, is the guarantee. Still prose-only; sealed contract untouched.

@greptile-apps review

PR #350, opened 2026-09-22
🔧 fix(config): show where each config list value comes from

### LINKED ISSUE #181: Initial developer experience: `tapes config list` is misleading before setup
## Summary

New contributors hit confusing signals on first run, even without running `tapes init`. The tool appears to be in a broken/uninitialized state when it's actually working fine with defaults.

## What's confusing

1. **`sqlite_path = <not set>`** doesn't indicate whether storage is disabled or falling back to a default. Sessions are in fact being written to `~/.tapes/tapes.sqlite`, but nothing in `tapes config list` says so.
2. **`~/.tapes/config.toml` exists without running `tapes init`.** Something on first run (likely `tapes serve` or the proxy) auto-creates a skeleton config, which contradicts the mental model of "I haven't initialized anything yet."
3. **The config file path is printed once at the top** of `tapes config list`, easy to miss while scanning key/value rows for the setting you care about.

## Repro

Fresh machine, no `tapes init`:
```
tapes serve api &         # or any first invocation that boots services
tapes config list
```
Observe: config file path exists, `storage.sqlite_path <not set>`, no indication where data is actually going.

## Suggestions

- Show **effective** values next to overrides, e.g. `storage.sqlite_path  <not set> (default: ~/.tapes/tapes.sqlite)`.
- Add a `tapes config doctor` (or extend `config list --verbose`) that prints the resolved storage path, DB size, and row counts so users can confirm "yes, data is landing here."
- Either don't auto-create `config.toml` until the user runs `init`, or print a one-line notice the first time it's created: `created ~/.tapes/config.toml with defaults`.

## Context

Caught by a new contributor during onboarding. Early-stage oversight — low severity, but it's the first impression for everyone joining.

aniruddhaadak80:
## Summary

`tapes config list` rendered effective values without saying which layer provided them, so built-in defaults looked explicitly configured (the confusion reported in #181). Each row now carries its source:

- `(default)` — built-in default, nothing sets this key
- `(config file)` — value comes from `config.toml`
- `(environment)` — value comes from a `TAPES_...` override

Keys with no value in any layer still show `<not set>`.

## What changed

- `pkg/config/source.go` (new): `ConfigSource` plus `Configer.GetConfigValueSource`, resolving value and source under the documented precedence (environment > config file > defaults). The shared viper setup moved into `loadViper`, reused by `GetConfigValue` with identical behavior.
- `cmd/tapes/config/list.go`: rows render the source annotation; long help documents the three markers.
- `pkg/config/source_test.go` (new): Ginkgo specs for default / config-file / environment sources, environment-over-file precedence, key-to-env mapping, and unknown-key rejection.
- `docs/configuration.md`: documents the annotations next to the `config list` example.

## Validation

- `go test ./pkg/config/ ./cmd/tapes/config/` — both suites pass (6 new specs included).
- `go vet ./pkg/config/ ./cmd/tapes/config/` — clean.
- Built `./cli/tapes` and ran `tapes config list` against an empty config dir: defaults show `(default)`, empty keys show `<not set>`; with a `config.toml` value and a `TAPES_...` override set, rows show `(config file)` and `(environment)` respectively.
- New specs fail without the fix (they reference the new `GetConfigValueSource` API).

Fixes #181


aniruddhaadak80:
Lint is green now. Root cause of the earlier tapes:check-lint error: the new exported type was named ConfigSource, which revive flags as stuttering (config.ConfigSource). Renamed to Source (config.Source); constants and call sites unchanged. Also rebased onto upstream/main 01f7561.

PR #351, opened 2026-09-22
🔧 fix(cli): drop the retired embed-worker from help output

aniruddhaadak80:
## Summary

The embed worker left core when search, skills, and export moved to their cassettes (#320), but two help strings still advertise it:

- the root `tapes --help` listed `tapes serve embed-worker` under "Run services", even though `serve` only has `api`, `derive-worker`, `ingest`, and `proxy` subcommands;
- `tapes serve derive-worker --help` told operators to "run the embed worker as a separate workload".

Both are copy-paste traps: following them fails.

## What changed

- `cmd/tapes/tapes.go`: removed the `tapes serve embed-worker` line from the root help.
- `cmd/tapes/serve/deriveworker/deriveworker.go`: the embedding paragraph now states that span embedding runs outside core in the search cassette.
- `cmd/tapes/help_test.go` (new): Ginkgo specs pinning the help surface — root help, the `serve` subcommand list, and the derive-worker help must not mention `embed-worker`.

## Validation

- `go test ./cmd/tapes/ ./cmd/tapes/serve/...` — all suites pass (3 new specs included); the new specs fail against the pre-fix help text.
- Built `./cli/tapes` before the fix: root `--help` showed the `embed-worker` line while `tapes serve --help` listed only `api`, `derive-worker`, `ingest`, `proxy`. After the fix, neither `--help` output mentions `embed-worker`.


aniruddhaadak80:
Lint is green now. Root cause of the earlier tapes:check-lint error: the new spec built the subcommand name list with make([]string, 0), which prealloc flags. Preallocated with capacity len(subs). Also rebased onto upstream/main 01f7561.

